PR #40, opened 2012-12-03
Background colour for the scene

stonier:
The rqt graph plugin shows up badly in dark coloured qt themes (black ellipses on a dark background). Usually it's just a mild annoyance, but found myself writing a similar plugin and caught the lines of code where the scene rendering is done in rqt graph, so here's a patch.

Note: I usually don't like forcing colours in an themed application, but in a rendered scene, your theme doesn't have a choice over the colours of the components (black ellipses for the nodes, etc), so it's choice of matching background is often wrong. This helps with that.


dirk-thomas:
This might as well apply to some other plugins. I will take a look soon...

PR #76, opened 2013-04-02
Use standard rospy function to filter ros remapping arguments from rqt plugin arguments

jbohren:
Currently, not all ROS remapping arguments are removed from python's sys.argv when passed to the rqt_gui plugins. This uses the builtin `rospy.myargs()` function to filter out these arguments.

### DIFF
--- a/rqt_gui/src/rqt_gui/main.py
+++ b/rqt_gui/src/rqt_gui/main.py
@@ -36,6 +36,8 @@ import argparse
 import os
 import sys
 
+import rospy
+
 from qt_gui.main import Main as Base
 
 
@@ -50,9 +52,7 @@ class Main(Base):
             argv = sys.argv
 
         # ignore ROS specific remapping arguments (see http://www.ros.org/wiki/Remapping%20Arguments)
-        remapping_arg_prefixes = ['__name', '__log', '__ip', '__hostname', '__master', '__ns']
-        remapping_arg_prefixes = ['%s:=' % arg for arg in remapping_arg_prefixes]
-        argv = [argv[0]] + [arg for arg in argv[1:] if not [p for p in remapping_arg_prefixes if arg.startswith(p)]]
+        argv = rospy.myargv(argv)
 
         return super(Main, self).main(argv, standalone=standalone, plugin_argument_provider=plugin_argument_provider)
 


jbohren:
Is there a reason why remapping arguments aren't allowed in rqt guis?

dirk-thomas:
Since there is only one ROS node the remapping arguments would apply for all plugins which is likely not desired.

jbohren:
Then if I want to remap the topics coming out of rqt_bag, or any other rqt_gui-based gui, that processing needs to happen in those modules, or is remapping just not going to be supported?

dirk-thomas:
ROS only supports remapping per-node. RQT runs a single ROS node (per language) (if not run in multi-process mode) and therefore can not provide per-plugin remapping.

dirk-thomas:
Passing the remapping arguments from the command line when invoking RQT can indeed be passed to the single ROS node (one per language) which is currently not done.

jbohren:
So the only feature you'd accept is a GUI remapping interface for rqt_bag?

dirk-thomas:
I am sorry but I can't follow you. I don't understand why it is about what I would accept nor what a GUI remapping interface for rqt_bag might be.

Please describe what your exact use case is - may be that help to clarify what feature you expect. We could also continue that discussion on the rqt mailing list ros-sig-rqt@googlegroups.com since it looks like it requires more information and discussion.

DorianScholz:
Any news on this issue?
Is there any chance to merge this back into Groovy?

dirk-thomas:
The hydro-devel branch only existed for a short amount of time without me knowing. Since there is not difference I have removed it some time ago. I will apply this patch to the groovy-devel branch. Thx for the comment.

130s:
@dirk-thomas 
For `hydro-devel` branch, I'm sure I created it when I was initial-releaseing into `hydro`, and removing the branch shouldn't matter at all.
Also, related to this, I added [branching policy to wiki](http://ros.org/wiki/rqt#Github_policy_for_rqt).

DorianScholz:
Thanks for applying the patch again. Any chance you could trigger a build to get this into the packages?

dirk-thomas:
A lot of core packages will be released in the next days. In this run I will also release rqt and its dependencies.

PR #81, opened 2013-06-05
rqt) Add stress on description about how to run rqt the easiest.

130s:


dirk-thomas:
This is stated on the wiki page already (directly under the first introduction paragraph). I don't think we need to replicate documentation in the package.xml file.

130s:
Ok. I thought that including important note in `package.xml` might be more effective since it comes at the top of wiki (I didn't mean to duplicate the same sentence), but I now come to think that notation on wiki makes this more readable.

Issue #67, opened 2013-01-31
inconsistent behavior on rqt_gui start without running roscore  (ros-pkg ticket #5562)


tfoote:
current behavior:
- roscpp plugins are available
- rospy plugins are disabled
  - rospy plugins that were loaded in the previous session are restored anyways, leading to init_node waiting for roscore
  - some plugins don't even need roscore (i.e. rqt_dep), but init_node is always called, when loading a plugin

desired behavior:
- same treatment for rospy and roscpp based plugins

option 1)
- plugins should specify in their plugin.xml whether they need roscore or not

option 2)
- rqt_gui should just launch a new roscore if none is running (nice for new users, maybe not desirable for advanced users)

trac data:
- Owner: **dthomas**
- Reporter: **dorianscholz**
- Reported at: **Tue Sep 18 04:34:55 2012**
- URL: https://code.ros.org/trac/ros-pkg/ticket/5562

dirk-thomas:
The exact same treatment is not possible as roscpp and rospy expose not the same functionality which would be necessary.

dirk-thomas:
This is addressed by #85

PR #85, opened 2013-08-19
delay node initialization for python plugins until start of plugins (#67)

dirk-thomas:
@ablasdel @DorianScholz What is your opinion on this? Is that better that how rqt currently behaves?

I could imagine one related feature: add a modal message box when a plugin start fails to make the message more visible. But that would require to collect the messages of multiple plugins when they are started together (either at start up or via a perspective) to not pop up numerous dialogs.

### DIFF
--- a/rqt_gui_cpp/src/rqt_gui_cpp/roscpp_plugin_provider.cpp
+++ b/rqt_gui_cpp/src/rqt_gui_cpp/roscpp_plugin_provider.cpp
@@ -41,6 +41,8 @@
 #include <pluginlib/class_list_macros.h>
 #include <ros/ros.h>
 
+#include <QMessageBox>
+
 #include <stdexcept>
 #include <sys/types.h>
 #include <unistd.h>
@@ -67,14 +69,31 @@ RosCppPluginProvider::~RosCppPluginProvider()
 
 void* RosCppPluginProvider::load(const QString& plugin_id, qt_gui_cpp::PluginContext* plugin_context)
 {
+  wait_for_master();
   init_node();
   return qt_gui_cpp::CompositePluginProvider::load(plugin_id, plugin_context);
 }
 
 qt_gui_cpp::Plugin* RosCppPluginProvider::load_plugin(const QString& plugin_id, qt_gui_cpp::PluginContext* plugin_context)
 {
+  wait_for_master();
   init_node();
   return qt_gui_cpp::CompositePluginProvider::load_plugin(plugin_id, plugin_context);
+}
+
+void RosCppPluginProvider::wait_for_master()
+{
+  while (!ros::master::check())
+  {
+    QMessageBox* mb = new QMessageBox(QMessageBox::Question, QObject::tr("Waiting for ROS master"), QObject::tr("Could not find ROS master. Either start a 'roscore' and retry or abort loading the plugin."), QMessageBox::Retry | QMessageBox::Abort);
+    mb->setDefaultButton(QMessageBox::Retry);
+    int button = mb->exec();
+    if (button == QMessageBox::Abort)
+    {
+      throw std::runtime_error("RosCppPluginProvider::init_node() could not find ROS master");
+    }
+    delete mb;
+  }
 }
 
 void RosCppPluginProvider::init_node()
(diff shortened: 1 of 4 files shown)


ablasdel:
I prefer this "failing on load" to the current behavior of not allowing people to select them. But I think we should probably have a better way of notifying the user about the issue than the stack trace that this results in. Is something supposed to be catching this exception? When I check it out nothing catches it.

dirk-thomas:
Can you please clarify what you mean with "failing on load". When the rqt application is started or when a specific plugin is loaded?

ablasdel:
The error is raised when the plugin is loaded. rqt remains up but there is a stack trace in the console and the plugin does not load. In my opinion this behavior is correct save for possibly the stack trace that it generates.

DorianScholz:
As Aaron stated, a back trace on the console is not very user friendly.
So just delaying it until the user starts a plugin does not help, but just postpones the problem.

How about we keep the check for the master on startup, but instead of disabling the (Python) plugins, ask the user in a modal dialog what to do: 
1) spawn a roscore for him
2) disable the plugins 
3) check for a master again: so the user can start roscore on its own and then press this button

And by the way, how is this handled for the C++ plugins? Consistency would be very nice...

ablasdel:
At the moment C plugins also throw an error. Just a different one. 
Python throws RuntimeError and C throws ImportError.

I think we should just catch both of these and have a recovery behavior like the modal dialog Dorian suggested but simpler. Something like "This plugin requires a roscore to run and we notice you don't have one running would you like rqt to start one for you? Yes/No" If they select no just kick them back without loading the plugin. No extra "disable plugin" states necessary.

This way it will be much clearer what is going on than disabling them since the user might not know why they are being disabled. Much nicer for users who don't know what is going on that is for sure.

dirk-thomas:
-1 on option 1). When you start nodes/launch files afterwards and have to restart your UI it breaks everything in a very subtle way.

-1 on option 2). Disabling is weird for the user. I agree with Aaron on that. And if a master is started afterwards it will not be reenabled again.

But I really like the idea of option 3). I just have to check how that behaves when restarting a perspective with multiple plugins open. But I guess it will just ask for the first and if the user starts a core and then confirms its fine. If he does not it might just ask again for every plugin. I will work on a pull request for that soon.

ablasdel:
+1 prompt the user to start a core if/when we need one and one isn't available.

dirk-thomas:
I have modified the pull request so that the master is checked before loading a plugin. If it is not available the user is asked how to proceed.

The second commit requires the pull request ros-visualization/qt_gui_core#25 as it uses a custom PluginLoadError exception to signal known cases of failures which will prevent the stacktrace in these cases.

DorianScholz:
@dirk-thomas I think you might have misunderstood my suggestion a bit.
I was talking of presenting the user with a messagebox which gives the user all the options 1), 2) and 3).
But you're right, 2) is a bit odd and might just as well be called "cancel loading plugin".
Nevertheless I like giving users option 1) for convenience, so they don't have to go to a terminal, start roscore and come back to click retry...
This should not interfere with any launchfile start / rqt restart, as the msgbox only comes up, if there is no master running.
So your current pull request fulfills this, except for option 1).

ablasdel:
The bulk of everything in this change is great and I +1 it with one reservation.
I think it would probably be better to have a simple ok modal dialog that states the error instead of the QWarning to the console. 

To talk a bit to the general case we might want to consider a general purpose dialog for errors, warnings and or unhandled exceptions instead of dumping them to the console. A nice dialog with the stack trace and as much info as possible would really make for a better user experience and give us more succinct and understandable information when we get issues reported.

dirk-thomas:
I do see the advantage for a user to be able to just click to start a roscore. But as I commented above I think that is a bad idea because when the users starts rqt, then some nodes and then restarts rqt it will tear down the master and restart a new one. Therefore rqt should never start a roscore. If that behvavior is desired we could add a launch file and just start rqt using roslaunch.

Regarding the general purpose dialog: you must be very careful to always spawn a modal dialog. It interrupt the workflow and blocks the user from going on. Only very severe events justify something like that. The missing roscore is arguably one but for any other that must be carefully judged on a case by case base. And generally I don't think that stacktraces belong in the UI at all.

ablasdel:
Ok ignoring the general case for the moment.

+1 modal dialog to inform the user of no roscore

DorianScholz:
@dirk-thomas 
Why would ending rqt tear down a roscore spawned from it?
Simply spawn the roscore in a new process and you're good. 
Try this minimal example if you like:

<pre>
#!/usr/bin/python
import subprocess
subprocess.Popen(['roscore > /dev/null'], shell=True)
</pre>

dirk-thomas:
With that approach you would loose the ability to tear the roscore down again. Try to describe to the potential numerous requests on answers.ros.org how to tear down this background process again :smile:

ablasdel:
I am going to -1 on the bringing up our own roscore. it isn't difficult for them to do separately and is not really worth the headaches it would to cause us to implement.

PR #87, opened 2013-09-02
Proposed replacement for rqt.svg

LucidOne:
A visually appealing upgrade of the rqt icon.


dirk-thomas:
"Visually appealing" might be subjective... ;-) But I think it is not a good thing to remove some of the dots for the "rqt" text since it does not follow the ROS style guide.

LucidOne:
How about if I remove the RQT and leave the monitor to make it more of an icon?

dirk-thomas:
I think that would make it better - and people should associate the icon with rqt so a label should not be necessary.

It might be an issue that the icon does not have any padding at the border. I think that was an issue when I added the current icon in the first place. If it fits in fine I will go ahead and merge the modified version after you committed it.

wjwwood:
@IHeartRobotics what about the icon in #89, thoughts?

wjwwood:
I exported it to a png with inkscape (hopefully it looks how you intended) for reference:

rqt

@spaepcke

dirk-thomas:
Closing in favor of #89.

PR #89, opened 2013-10-09
new icon for rqt

wjwwood:
The flat, square background did not look good on OS X. It is still blurry, but there seems to be no way to get hidpi images in the dock without using a .app bundle and an info.plist file.

I created it with omnigraffle to get the drop shadows and stuff, so I saved the graffle file into the repo as well. I exported a svg copy of it, but it seems to use svg stuff that Qt doesn't support, plus it doesn't look any better. So I saved out a 1024x1024 png, which looks fine and has the drop shadow. I haven't seen what this looks like on Linux.

This is the png:

rqt

This is what it looks like in my dock:

screen shot 2013-10-08 at 9 28 29 pm

This is what the old one looked like:

screen shot 2013-10-08 at 9 36 39 pm

I tried to make the new one using svg directly, but I couldn't get the rounded edges to behave. In the end it is sort of subjective, but I personally think this looks nicer, thoughts?


wjwwood:
@dirk-thomas @ablasdel @spaepcke

wjwwood:
This is what the switching looks like (had to figure out how to capture a screenshot with cmd-tab going):

rqt_icon_ss

And the old one:

rqt_old_icon_ss

ablasdel:
+1 I much prefer the sleek rounded look of your new icon.

dirk-thomas:
Have you seen #87?

spaepcke:
I wasn't able to see the image in #87. Screenshot available?

William, I like the rounded corner version you sent.

dirk-thomas:
Please update the CMakeLists.txt to install the rqt.png instead of the rqt.svg. And update the pr to apply to latest version.

wjwwood:
Will do, has anyone checked it out on Linux, does it look ok? Could you post a screenshot?

PR #140, opened 2018-10-19
fixing tests so that they can run locally

mlautman:


rqt_py_common/CMakeLists.txt
```diff
@@ -22,5 +22,7 @@ install(DIRECTORY resource
 )
 
 if(CATKIN_ENABLE_TESTING)
-  catkin_add_nosetests(test)
+  catkin_add_nosetests(test/test_rqt_common_unit.py)
+  catkin_add_nosetests(test/test_rqt_ros_graph.py)
+  catkin_add_nosetests(test/test_rqt_roscomm_util.py)
```
dirk-thomas, line 26: Same as https://github.com/ros-visualization/qt_gui_core/pull/123#discussion_r226803085

rqt_py_common/package.xml
```diff
@@ -37,6 +37,8 @@
   <run_depend>actionlib</run_depend>
   <run_depend>rosbag</run_depend>
 
+  <test_depend>pr2_moveit_config</test_depend>
```
dirk-thomas: Instead of adding this test dependency I would suggest to modify the test to check for a more "common" package. Otherwise this package can't be released into a new distro until `pr2_moveit_config` has been released which is a huge restriction.

rqt_py_common/test/test_rqt_ros_graph.py
```diff
@@ -82,9 +82,11 @@ def test_get_lower_grn_dfs(self):
                                                   '')),
                          self._len_lower_grn_node1_1)
 
-    def test_get_full_grn(self):
-        self.assertEqual(RqtRosGraph.get_full_grn(self._node1_1_1.index()),
-                         self._grn_node1_1_1)
+    # TODO(mlautman): get_full_grn is now depricated thus there is no reason
```
dirk-thomas: Spelling: "depricated"

rqt_py_common/CMakeLists.txt
mlautman: @dirk-thomas I'm not sure that nosetest is working as expected here. I dug into the catkin_tools code but it became clear that it would lead to serious scope creep. This seems like the most straightforward way to fix the issue and I'm inclined to leave it unless you strongly object

dirk-thomas: The actual problem is that two of the test files have the executable bit set. Nosetest won't pick executables. So removing the executable bit will allow to keep passing the directory name.

mlautman: This seems like it needs to be documented somewhere..

mlautman: This did not fix the issue...

mlautman: @dirk-thomas Can you elaborate on what you mean by 'two files have the executable bit set'? I deleted out the main from the test files and found the same issue.

dirk-thomas: > This seems like it needs to be documented somewhere.

It is documented: https://nose.readthedocs.io/en/latest/usage.html#extended-usage

> It is important to note that the default behavior of nose is to not include tests from files which are executable.

The two files `test_rqt_roscomm_util.py` and `test_rqt_ros_graph.py` need the executable bit removed (`chmod a-x ...`).

mlautman:
Stale

PR #142, opened 2018-10-30
Porting rqt_gui to ros2

brawner:
This ports the rqt_gui package from rqt. For the most part it changes rospack logic to ament_index logic and creates an ament package.

There is one todo added here, which requires parsing any arguments with ros logic. Before this was handled with rospy, but there doesn't appear to be a specific hook with rclpy yet.

This also incorporates the style changes in #143. So for comparison sake that one might need to be merged first.

### DIFF
--- a/rqt_gui/CMakeLists.txt
+++ b/rqt_gui/CMakeLists.txt
@@ -1,13 +1,18 @@
-cmake_minimum_required(VERSION 2.8.3)
+cmake_minimum_required(VERSION 3.5)
 project(rqt_gui)
+
 # Load catkin and all dependencies required for this package
-find_package(catkin REQUIRED COMPONENTS qt_gui)
-catkin_package()
-catkin_python_setup()
+find_package(ament_cmake REQUIRED)
+find_package(ament_cmake_python REQUIRED)
+
+ament_python_install_package(${PROJECT_NAME}
+  PACKAGE_DIR src/${PROJECT_NAME})
 
 install(PROGRAMS bin/rqt_gui
-  DESTINATION ${CATKIN_PACKAGE_BIN_DESTINATION}
+  DESTINATION bin
 )
-install(FILES resource/rqt.png
-  DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION}/resource
+install(DIRECTORY resource
+  DESTINATION share
 )
+
+ament_package()
--- a/rqt_gui/package.xml
+++ b/rqt_gui/package.xml
@@ -1,4 +1,4 @@
-<package>
+<package format="2">
   <name>rqt_gui</name>
   <version>0.5.0</version>
   <description>rqt_gui provides the main to start an instance of the ROS integrated graphical user interface provided by qt_gui.</description>
@@ -12,9 +12,18 @@
 
   <author>Dirk Thomas</author>
 
-  <buildtool_depend>catkin</buildtool_depend>
+  <buildtool_depend>ament_cmake</buildtool_depend>
 
   <build_depend version_gte="0.3.0">qt_gui</build_depend>
-  <run_depend>catkin</run_depend>
-  <run_depend version_gte="0.3.0">qt_gui</run_depend>
+
+  <exec_depend>ament_cmake_python</exec_depend>
+  <exec_depend>python_qt_binding</exec_depend>
+  <exec_depend>python3-catkin-pkg-modules</exec_depend>
+  <exec_depend>qt_gui</exec_depend>
+  <exec_depend>rclpy</exec_depend>
+  <exec_depend version_gte="0.3.0">qt_gui</exec_depend>
+
+  <export>
+    <build_type>ament_cmake</build_type>
+  </export>
 </package>
--- a/rqt_gui/setup.py
+++ b/rqt_gui/setup.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python
 
 from distutils.core import setup
+
 from catkin_pkg.python_setup import generate_distutils_setup
 
 d = generate_distutils_setup(
--- a/rqt_gui/setup.py
+++ b/rqt_gui/setup.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 from distutils.core import setup
 
@@ -7,7 +7,7 @@ from catkin_pkg.python_setup import generate_distutils_setup
 package_name = "rqt_gui"
 
 d = generate_distutils_setup(
-    packages=['rqt_gui'],
+    packages=[package_name],
     package_dir={'': 'src'},
     scripts=['bin/rqt'],
     data_files=[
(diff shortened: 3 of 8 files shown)


rqt_gui/package.xml
```diff
@@ -1,4 +1,4 @@
-<package>
+<package format="2">
   <name>rqt_gui</name>
   <version>0.5.0</version>
```
mlautman, line 3: version bump? @dirk-thomas can you weigh in on how you envision updating version numbers when there are separate ROS1 and ROS2 packages being released?

rqt_gui/setup.py
```diff
@@ -1,12 +1,20 @@
 #!/usr/bin/env python
```
mlautman: python3

rqt_gui/setup.py
```diff
@@ -1,12 +1,20 @@
...
 
 from distutils.core import setup
+
 from catkin_pkg.python_setup import generate_distutils_setup
 
+package_name = "rqt_gui"
+
 d = generate_distutils_setup(
     packages=['rqt_gui'],
```
mlautman: use `package_name` here

rqt_gui/src/rqt_gui/main.py
```diff
@@ -35,37 +35,34 @@
...
 
     def main(self, argv=None, standalone=None, plugin_argument_provider=None):
         if argv is None:
             argv = sys.argv
 
-        # ignore ROS specific remapping arguments (see
-        # http://www.ros.org/wiki/Remapping%20Arguments)
-        argv = rospy.myargv(argv)
+        # ignore ROS specific remapping arguments
+        # TODO port to ros2
```
mlautman: add your github handle `TODO(brawner): blah blah` and add some more information why this is being left for the future

rqt_gui/setup.py
```diff
@@ -1,12 +1,20 @@
...
     package_dir={'': 'src'},
-    scripts=['bin/rqt']
+    scripts=['bin/rqt'],
+    data_files=[
+        ('share/ament_index/resource_index/packages',
+            ['resource/' + package_name]),
+        ('share/' + package_name, ['package.xml']),
```
mlautman, line 16: do you need to add the package.xml to the shared data files? I don't understand what this is doing..

mlautman (commented):
+1 after addressing comments

rqt_gui/package.xml
dirk-thomas: Since the tags from different ROS distros (1 and 2) will share the same namespace we need room to grow the ROS 1 package versions in future releases. Therefore I would suggest using 1.0.0 for the ROS 2 branch.

brawner: Changed to 1.0.0

rqt_gui/setup.py
brawner: I'm following the lead of demo_nodes_py 
https://github.com/ros2/demos/blob/master/demo_nodes_py/setup.py

Since this is an ament_cmake package I don't think setup.py is run. But this copies the package.xml to the package's share directory in install.

rqt_gui/CMakeLists.txt
```diff
@@ -1,13 +1,18 @@
...
+  PACKAGE_DIR src/${PROJECT_NAME})
 
 install(PROGRAMS bin/rqt_gui
-  DESTINATION ${CATKIN_PACKAGE_BIN_DESTINATION}
+  DESTINATION bin
 )
-install(FILES resource/rqt.png
-  DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION}/resource
+install(DIRECTORY resource
+  DESTINATION share
```
dirk-thomas, line 15: Needs to install into package specific directory.

rqt_gui/package.xml
```diff
@@ -12,9 +12,18 @@
...
   <author>Dirk Thomas</author>
 
-  <buildtool_depend>catkin</buildtool_depend>
+  <buildtool_depend>ament_cmake</buildtool_depend>
 
   <build_depend version_gte="0.3.0">qt_gui</build_depend>
-  <run_depend>catkin</run_depend>
-  <run_depend version_gte="0.3.0">qt_gui</run_depend>
+
+  <exec_depend>ament_cmake_python</exec_depend>
```
dirk-thomas, line 19: Why is this an `exec_depend`?

rqt_gui/package.xml
```diff
@@ -12,9 +12,18 @@
...
+  <exec_depend>python_qt_binding</exec_depend>
+  <exec_depend>python3-catkin-pkg-modules</exec_depend>
+  <exec_depend>qt_gui</exec_depend>
```
dirk-thomas, line 22: Duplicate of line 24?

rqt_gui/setup.py
```diff
@@ -1,13 +1,20 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
```
dirk-thomas, line 1: This whole file isn't used since the package uses the build type `ament_cmake`.

rqt_gui/src/rqt_gui/main.py
```diff
@@ -36,38 +36,37 @@
...
 
     def main(self, argv=None, standalone=None, plugin_argument_provider=None):
         if argv is None:
             argv = sys.argv
 
-        # ignore ROS specific remapping arguments (see
-        # http://www.ros.org/wiki/Remapping%20Arguments)
-        argv = rospy.myargv(argv)
+        # ignore ROS specific remapping arguments
+        # TODO(brawner) port to ros2. Waiting on a feature add and PR on rclpy
```
dirk-thomas, line 57: Please create an issue for this to follow up on.

rqt_gui/src/rqt_gui/main.py
```diff
@@ -36,38 +36,37 @@
...
-                             plugin_manager_settings_prefix=str(
-                                hash(os.environ['ROS_PACKAGE_PATH'])))
+                             plugin_argument_provider=plugin_argument_provider)
 
     def create_application(self, argv):
         from python_qt_binding.QtGui import QIcon
         app = super(Main, self).create_application(argv)
-        logo = os.path.join(self._ros_pack.get_path('rqt_gui'), 'resource', 'rqt.png')
+        rqt_gui_path = get_package_path('rqt_gui')
+        logo = os.path.join(rqt_gui_path, 'share', 'resource', 'rqt.png')
```
dirk-thomas, line 69: Needs to be updated once the install destination has been fixed.

rqt_gui/package.xml
brawner: Looks like it was supposed to be ament_index_python

rqt_gui/CMakeLists.txt
dirk-thomas: Addressed by #153.

rqt_gui/setup.py
dirk-thomas: Addressed by #153.

rqt_gui/src/rqt_gui/main.py
dirk-thomas: Addressed by #153.

rqt_gui/package.xml
dirk-thomas: Addressed by #153.

PR #143, opened 2018-10-30
Style changes

brawner:
This introduces a number of style changes to be compatible with ROS2 style guidelines, mostly centered around line lengths and documentation style.

### DIFF
--- a/rqt_gui/setup.py
+++ b/rqt_gui/setup.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python
 
 from distutils.core import setup
+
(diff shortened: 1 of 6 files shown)


dirk-thomas (approved):

dirk-thomas:
Should #145 be integrated to only do one style change commit?

brawner:
@dirk-thomas I thought I would've made the PRs smaller this time so they would be easier to review and manage.

I pulled in #145 to this PR so now changes to rqt_gui and rqt_gui_py are here.

mlautman (approved):

PR #145, opened 2018-10-30
Style changes rqt gui py

brawner:
This PR implements style changes for ROS2 compatibility. Primarily, it fixes a few line lengths that were not caught previously.

### DIFF
--- a/rqt_gui_py/src/rqt_gui_py/plugin.py
+++ b/rqt_gui_py/src/rqt_gui_py/plugin.py
@@ -34,12 +34,13 @@ from qt_gui.plugin import Plugin as Base
 
 
 class Plugin(Base):
-
     """
     Interface for Python plugins which use the ROS client library.
-    User-defined plugins may either subclass `rqt_gui_py.plugin.Plugin` or according to duck typing implement only the needed methods.
-    A plugin must not call rospy.init_node() as this is performed once by the framework.
-    The name of the ROS node consists of the prefix "rqt_gui_py_node_" and the process id.
+
+    User-defined plugins may either subclass `rqt_gui_py.plugin.Plugin` or according to duck typing
+    implement only the needed methods. A plugin must not call rospy.init_node() as this is
+    performed once by the framework. The name of the ROS node consists of the prefix
+    "rqt_gui_py_node_" and the process id.
     """
 
     def __init__(self, context):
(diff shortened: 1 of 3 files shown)


rqt_gui_py/src/rqt_gui_py/plugin.py
```diff
@@ -34,12 +34,13 @@
...
     """
     Interface for Python plugins which use the ROS client library.
-    User-defined plugins may either subclass `rqt_gui_py.plugin.Plugin` or according to duck typing implement only the needed methods.
-    A plugin must not call rospy.init_node() as this is performed once by the framework.
-    The name of the ROS node consists of the prefix "rqt_gui_py_node_" and the process id.
+
+    User-defined plugins may either subclass `rqt_gui_py.plugin.Plugin` or according to duck typing
+    implement only the needed methods. A plugin must not call rospy.init_node() as this is
+    performed once by the framework. The name of the ROS node consists of the prefix
+    "rqt_gui_py_node_" and the process id.
```
dirk-thomas: Each sentence should still start on a new line.

rqt_gui_py/src/rqt_gui_py/ros_py_plugin_provider.py
```diff
@@ -72,8 +72,9 @@ def _check_for_master(self):
         self._wait_for_master_thread.start()
         self._wait_for_master_dialog = QMessageBox(QMessageBox.Question,
                                                    self.tr('Waiting for ROS master'),
-                                                   self.tr("Could not find ROS master. Either start a 'roscore' or "
-                                                           "abort loading the plugin."),
+                                                   self.tr("Could not find ROS master. Either "
+                                                           "start a 'roscore' or abort loading the"
+                                                           "plugin."),
```
dirk-thomas: This is a good example of why aligning the content with the open parenthesis is a "bad" idea. Due to the massive indentation the usable line length is tiny which results in several wraps and hard to read content. Please wrap after `QMessageBox(` and indent the block with just 4 spaces. Also the reformatting has removed the space between the last two words "the" and "plugin".

dirk-thomas (commented):
Can this be rolled into #143?

brawner:
After addressing @dirk-thomas's comments, I'm merging this into #143

PR #147, opened 2018-11-03
Porting rqt_gui_py to ROS2 with testing

brawner:
This ports the package rqt_gui_py to ROS2 with flake8 testing included. 

Passing CI build https://ci.ros2.org/job/ci_linux/5458

### DIFF
--- a/rqt_gui_py/src/rqt_gui_py/ros_py_plugin_provider.py
+++ b/rqt_gui_py/src/rqt_gui_py/ros_py_plugin_provider.py
@@ -29,23 +29,15 @@
 # POSSIBILITY OF SUCH DAMAGE.
 
 import os
-import threading
-import time
+
+from python_qt_binding.QtCore import qDebug
+from qt_gui.composite_plugin_provider import CompositePluginProvider
 
 import rclpy
-
-from qt_gui.composite_plugin_provider import CompositePluginProvider
-from qt_gui.errors import PluginLoadError
-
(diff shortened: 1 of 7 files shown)


rqt_gui_py/src/rqt_gui_py/ros_py_plugin_provider.py
```diff
@@ -29,77 +29,32 @@
...
-            break
-
     def _init_node(self):
         # initialize node once
         if not self._node_initialized:
             name = 'rqt_gui_py_node_%d' % os.getpid()
             qDebug('RosPyPluginProvider._init_node() initialize ROS node "%s"' % name)
-            rospy.init_node(name, disable_signals=True)
+            rclpy.init()
+            self._node = rclpy.create_node(name)
```
mlautman, line 63: Is this called automagically or does the user need to call _init_node? 

Also, you may want to make explicit destroying the node an rclpy.shutdown() in a __del__(self) method

mlautman (commented):
+1 after addressing feedback

rqt_gui_py/src/rqt_gui_py/ros_py_plugin_provider.py
brawner: _init_node() is called from the public method load() above. I can't actually find a plugin yet that implements this class.

I'll add similar unload and _destroy_node methods.

PR #148, opened 2018-11-03
rqt_py_common port

mlautman:
Porting rqt_py_common to ROS2 crystal. 
The tests will not pass without https://github.com/ros2/rosidl_python/pull/19


rqt_py_common/package.xml
```diff
@@ -19,25 +19,27 @@
...
 
-  <run_depend>genpy</run_depend>
-  <run_depend version_gte="0.2.19">python_qt_binding</run_depend>
-  <run_depend>qt_gui</run_depend>
-  <run_depend>roslib</run_depend>
-  <run_depend>rospy</run_depend>
-  <run_depend>rostopic</run_depend>
+  <exec_depend>rosidl_default_runtime</exec_depend>
+  <exec_depend version_gte="0.3.4">python_qt_binding</exec_depend>
+  <exec_depend>qt_gui</exec_depend>
```
brawner, line 30: Depends should be in alphabetical order

rqt_py_common/package.xml
```diff
@@ -19,25 +19,27 @@
...
+  <test_depend>ament_lint_auto</test_depend>
+  <test_depend>ament_lint_common</test_depend>
+  <test_depend>ament_cmake_pytest</test_depend>
+  <test_depend>ament_flake8</test_depend>
+  <test_depend>ament_pep257</test_depend>
+  <test_depend>python3-pytest</test_depend>
 
   <export>
+    <build_type>ament_cmake</build_type>
     <architecture_independent/>
```
brawner, line 44: Does the <architecture_independent/> export exist in ROS2? I've been removing them

rqt_py_common/setup.py
```diff
@@ -1,11 +1,8 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 from distutils.core import setup
-from catkin_pkg.python_setup import generate_distutils_setup
```
brawner, line 4: FYI: I don't think this needs to be switched from catkin_pkg to distutils right? Both seem to work fine. But no need to change it back

rqt_py_common/src/rqt_py_common/plugin_container_widget.py
```diff
@@ -104,7 +106,7 @@ def set_sysmsg(self, sysmsg):
         Set system msg that's supposed to be shown in sys msg pane.
         @type sysmsg: str
         """
-        rospy.loginfo('PluginContainerWidget; {}'.format(sysmsg))
+        self._logger.info('{}'.format(sysmsg))
```
brawner, line 111: Does ros2 logger display the context of the info message? Like which file this is being logged from?

rqt_py_common/src/rqt_py_common/plugin_container_widget.py
```diff
@@ -75,8 +74,11 @@ def __init__(self, plugin_widget,
         """
         super(PluginContainerWidget, self).__init__()
 
-        ui_file = os.path.join(rospkg.RosPack().get_path('rqt_py_common'),
-                               'resource', 'plugin_container.ui')
+        self._logger = rclpy.logging.get_logger("PluginContainerWidget")
+        _, package_path = get_resource('packages', 'rqt_py_common')
```
brawner, line 80: This could be replaced with a call to qt_gui.ros_package_helper.get_package_path()

rqt_py_common/src/rqt_py_common/rqt_roscomm_util.py
```diff
@@ -34,76 +34,83 @@
...
+        #         "load_parameters: unable to set params (last param was [%s]): %s" % (param, e))
+        #     raise  # re-raise as this is fatal
+        # RqtRoscommUtil._logger.debug("... load_parameters complete")
+        RqtRoscommUtil._logger.error("load_parameters: not implemented in ROS2))")
+        pass
 
     @staticmethod
     def iterate_packages(subdir):
         """
+        Note: Mlautman 11/2/2018
```
brawner: Is the high level idea of this code is to get packages that contain messages/services/actions etc? Because there are some examples of that in the rcl command line package. For example, to discover packages that contain message definitions

https://github.com/ros2/ros2cli/blob/211325816e03baaa390c65551b17dece5495c509/ros2msg/ros2msg/api/__init__.py#L22

rqt_py_common/src/rqt_py_common/topic_helpers.py
```diff
@@ -28,161 +28,243 @@
 # ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 # POSSIBILITY OF SUCH DAMAGE.
 
-import roslib
-import roslib.msgs
-import roslib.message
-from rostopic import get_topic_type
-from python_qt_binding.QtCore import qDebug
+import rclpy
```
brawner: These imports should be in alphabetical order

rqt_py_common/src/rqt_py_common/topic_helpers.py
```diff
@@ -28,161 +28,243 @@
...
+    topic_list = node.get_topic_names_and_types()
+
+    node.destroy_node()
+    if shutdown_rclpy:
+        rclpy.shutdown()
+
+    return topic_list
+
+
+_logger = logging.get_logger("topic_helpers")
```
brawner: Maybe instead of creating a global variable, you just use
logging.get_logger().info()

rqt_py_common/src/rqt_py_common/rqt_roscomm_util.py
```diff
@@ -199,12 +213,14 @@ def mfilter(f):
 
     @staticmethod
     def is_roscore_running():
```
brawner: Should functions like these be deprecated or just removed?

rqt_py_common/src/rqt_py_common/rqt_roscomm_util.py
```diff
@@ -112,22 +119,28 @@ def iterate_packages(subdir):
...
+        #     d = os.path.join(rospack.get_path(p), subdir)
+        #     RqtRoscommUtil._logger.debug('rospack dir={}'.format(d))
+        #     if os.path.isdir(d):
+        #         yield p, d
+        RqtRoscommUtil._logger.error("iterate_packages: not implemented in ROS2))")
 
     @staticmethod
     def list_files(package, subdir, file_extension='.launch'):
         """
+        Note: Mlautman 11/2/2018
```
brawner, line 133: Shouldn't there be a function like this to find launch files and any other necessary ros2 files for a package-specific subdirectory?

brawner (changes requested):
Good work. Looks like it was a large package to port over. I think a few functions that you are deprecating should be changed or modified to fit the ros2 paradigm.

mlautman:
test:
https://ci.ros2.org/job/ci_linux/5476/

rqt_py_common/package.xml
mlautman: I have no idea. @dirk-thomas?

rqt_py_common/src/rqt_py_common/plugin_container_widget.py
mlautman: `from rclpy import logging`
`_logger = logging.get_logger("topic_helpers")`
`_logger.info("yo")`

will print out:
`[INFO] [topic_helpers]: yo`

rqt_py_common/src/rqt_py_common/plugin_container_widget.py
mlautman: It could but unless you feel strongly about this I'd rather not change it.

rqt_py_common/src/rqt_py_common/rqt_roscomm_util.py
mlautman: I've been leaving the ones that I think may be ported in the future. The rest can be deleted

rqt_py_common/src/rqt_py_common/rqt_roscomm_util.py
mlautman: implemented

rqt_py_common/src/rqt_py_common/rqt_roscomm_util.py
mlautman: whoa that's cool, thanks for sharing!

rqt_py_common/src/rqt_py_common/topic_helpers.py
mlautman: addressed

rqt_py_common/src/rqt_py_common/topic_helpers.py
mlautman: Seems low priority but I did it anyway

rqt_py_common/src/rqt_py_common/extended_combo_box.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
```
mlautman, line 1: @dirk-thomas Can you confirm that these lines can be deleted for all the files that aren't executable

rqt_py_common/src/rqt_py_common/topic_helpers.py
```diff
@@ -28,161 +28,168 @@
...
-            return bool
-        else:
-            return type(roslib.msgs._convert_val(type_name, 0))
-    else:
-        return roslib.message.get_message_class(type_name)
+from rclpy import logging
 
 
-def get_field_type(topic_name):
+def get_topic_names_and_types(node=None):
```
mlautman, line 38: @dirk-thomas This feels like a bad way to do this, initialize rclpy and create a node just to get the topic names and types but I don't know of a better way... your thoughts?

rqt_py_common/package.xml
mlautman: done

rqt_py_common/setup.py
mlautman: @dirk-thomas?

rqt_py_common/src/rqt_py_common/topic_helpers.py
brawner: Here's an example from ros2cli DirectNode. Creates and destroys a temporary node. It includes _enter_ and _exit_ methods so you can use with syntax.

https://github.com/ros2/ros2cli/blob/master/ros2cli/ros2cli/node/direct.py

mlautman: very cool

rqt_py_common/CMakeLists.txt
```diff
@@ -1,26 +1,33 @@
...
 install(DIRECTORY resource
-  DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION}
+  DESTINATION share/${PROJECT_NAME}
 )
 
-if(CATKIN_ENABLE_TESTING)
-  catkin_add_nosetests(test)
-endif()
\ No newline at end of file
+install(TARGETS ${_library_name}
```
dirk-thomas, line 15: Where is the variable `_library_name` being defined?

rqt_py_common/package.xml
```diff
@@ -19,25 +19,28 @@
...
-  <!-- 3/11/2013 depending on actionlib & rosbag is only temporary;
-  rosaction.py needs it. In the future this run_depend will be removed when
-  rosaction gets moved to actionlib. -->
-  <run_depend>actionlib</run_depend>
-  <run_depend>rosbag</run_depend>
+  <test_depend>ament_lint_auto</test_depend>
+  <test_depend>ament_lint_common</test_depend>
+  <test_depend>ament_cmake_pytest</test_depend>
+  <test_depend>ament_flake8</test_depend>
+  <test_depend>ament_pep257</test_depend>
```
dirk-thomas, line 37: Same comment as on [qt_gui_core](https://github.com/ros-visualization/qt_gui_core/pull/133#pullrequestreview-172627653).

rqt_py_common/setup.py
```diff
@@ -1,11 +1,8 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
```
dirk-thomas, line 1: Same comment as on [qt_gui_core](https://github.com/ros-visualization/rqt/pull/142/files/3fee97e14908cb061c8977ac5609f1fce965ce32#diff-1ab904ac4fc501b59ea1d8c30ca50ec3).

rqt_py_common/src/rqt_py_common/layout_util.py
```diff
@@ -34,11 +34,12 @@
...
 from python_qt_binding.QtCore import Qt
 from python_qt_binding.QtWidgets import QWidgetItem, QSpacerItem
-import roslib
-import rospy
+
+from rclpy import logging
 
 
 class LayoutUtil(object):
+    _logger = logging.get_logger("LayoutUtil")
```
dirk-thomas, line 42: Should use single quotes.

Same below.

rqt_py_common/src/rqt_py_common/rosaction.py
```diff
@@ -1,884 +0,0 @@
...
-
-def rosaction_debug(rospack, mode, type_, raw=False):
-    """
-    Prints contents of msg/srv file
-    :param mode: MODE_ACTION or MODE_SRV, ``str``
-    """
-    if mode == MODE_ACTION:
-        print(get_msg_text(type_, raw=raw, rospack=rospack))
-    else:
-        raise ROSActionException("Invalid mode for debug: %s" % mode)
```
dirk-thomas, line 884: Removing this file and adding it later again will "loose" the history and is very unfortunate.

rqt_py_common/src/rqt_py_common/rqt_roscomm_util.py
```diff
@@ -34,72 +34,76 @@
...
 
 class RqtRoscommUtil(object):
+    _logger = logging.get_logger("RqtRoscommUtil")
 
     @staticmethod
     def load_parameters(config, caller_id):
         """
+        NOTE: Mlautman 11/2/18
+              This function has been deprecated as ROS2 does not yet support setting
+              and getting of parameters in Python
```
dirk-thomas, line 50: I am not sure what the idea of this method is but `rclpy` does have support for parameters.

rqt_py_common/src/rqt_py_common/rqt_roscomm_util.py
```diff
@@ -137,14 +144,8 @@ def list_files(package, subdir, file_extension='.launch'):
...
         """
-        if subdir == None or subdir == '':
-            raise ValueError('Invalid package subdir = {}'.format(subdir))
-
-        rospack = rospkg.RosPack()
-
-        path = os.path.join(rospack.get_path(package), subdir)
-
-        return [genmsg.resource_name(package, t) for t in RqtRoscommUtil._list_types(path, file_extension)]
+        RqtRoscommUtil._logger.error("list_files: not implemented in ROS2))")
```
dirk-thomas, line 147: Please create a ticket for this to follow up on.

rqt_py_common/src/rqt_py_common/topic_helpers.py
```diff
@@ -28,161 +28,167 @@
...
+    if not rclpy.ok():
+        shutdown_rclpy = True
+        rclpy.init()
 
-    slot_path = topic_name[len(real_topic_name):]
-    return get_slot_type(message_class, slot_path)
+    node = rclpy.create_node("TopicHelpers__get_topic_names_and_types")
 
+    # Give the node time to learn about the graph
+    rclpy.spin_once(node, timeout_sec=0.05)
```
dirk-thomas, line 61: This looks pretty fragile. This timeout is very likely not enough - and no hard coded value will be good for all cases.

rqt_py_common/src/rqt_py_common/topic_helpers.py
```diff
@@ -28,161 +28,167 @@
...
-    for field_name in fields:
+    node.destroy_node()
+    if shutdown_rclpy:
+        rclpy.shutdown()
+
+    return topic_list
+
+
+_message_class_cache = {}
+def get_message_class(message_type):
```
dirk-thomas, line 73: The logic in this function is certainly going to break with the upcoming IDL changes.

rqt_py_common/package.xml
dirk-thomas: Addressed by #153.

rqt_py_common/src/rqt_py_common/layout_util.py
mlautman: Seems subjective. My understanding is that single and double quotes are interchangeable in python

rqt_py_common/src/rqt_py_common/rosaction.py
mlautman: According to the todo's this file needs to be moved to actionlib. I didn't know it was possible to move a file across repos and preserve history

rqt_py_common/src/rqt_py_common/rqt_roscomm_util.py
mlautman: I was basing this off the wiki on ros2 that states that bouncy does not have support parameters in python yet. Can you point me to an implementation?

mlautman: Also, this method isn't used anywhere in the RQT packages being ported to ROS2. Is there a reason we wouldn't want to remove it?

rqt_py_common/src/rqt_py_common/rqt_roscomm_util.py
mlautman: https://github.com/ros-visualization/rqt/issues/154

rqt_py_common/src/rqt_py_common/topic_helpers.py
mlautman: Is there a better way to do this? I modeled this after https://github.com/ros2/ros2cli/blob/master/ros2cli/ros2cli/node/direct.py

rqt_py_common/src/rqt_py_common/topic_helpers.py
mlautman: Definitely ping me when the upcoming idl changes are solidified

rqt_py_common/src/rqt_py_common/rqt_roscomm_util.py
dirk-thomas: > I was basing this off the wiki on ros2 that states that bouncy does not have support parameters in python yet.

Python parameters have been added post Bouncy. Since we target the Crystal release we can use them.

> this method isn't used anywhere in the RQT packages being ported to ROS2.

You could follow the history to see when and why it was added.

> Is there a reason we wouldn't want to remove it?

Assuming it was added for a reason and some plugins depend on it if we don't port it now it likely needs to happen some time later.

rqt_py_common/CMakeLists.txt
brawner: Resolved in #155

rqt_py_common/setup.py
brawner: Resolved in #155

rqt_py_common/src/rqt_py_common/topic_helpers.py
brawner: Addressed in #155

rqt_py_common/src/rqt_py_common/rqt_roscomm_util.py
dirk-thomas: The same not accurate note is still present in #155.

rqt_py_common/src/rqt_py_common/topic_helpers.py
dirk-thomas: You mentioned that the problem has been addressed but the referenced PR just uses a higher hard coded value. I don't think that is sufficient. And I don't mean the number should be higher but this does conceptionally not address the problem.

mlautman: @dirk-thomas I have a separate PR for using the daemon approach that you detailed. Also, @brawner's PR https://github.com/ros-visualization/rqt/pull/161 provides a better fix that we will want to go back and use throughout the code base.

rqt_py_common/src/rqt_py_common/rqt_roscomm_util.py
mlautman: I would like to take care of this in a future PR. As of now, #155 is too big to keep adding features

PR #152, opened 2018-11-07
Updating rqt_gui CMakeLists to properly install rqt_gui executable

brawner:
This moves the rqt_gui python executable to the lib directory, where it can be found as an executable.


dirk-thomas:
:+1:

PR #155, opened 2018-11-08
addressing Dirk's feedback on the rqt_py_common port

mlautman:
This cleans up some rqt_py_common code and adds the get_field_type function call

### DIFF
--- a/rqt_py_common/CMakeLists.txt
+++ b/rqt_py_common/CMakeLists.txt
@@ -27,6 +27,9 @@ if(BUILD_TESTING)
     test/msg/Val.msg
     test/msg/ArrayVal.msg
     DEPENDENCIES std_msgs)
+  set(ament_cmake_copyright_FOUND TRUE)
+  ament_lint_auto_find_test_dependencies()
+
   ament_add_pytest_test(${PROJECT_NAME} test
     APPEND_ENV PYTHONPATH=${CMAKE_CURRENT_BINARY_DIR}
     TIMEOUT 90)
--- a/rqt_py_common/CMakeLists.txt
+++ b/rqt_py_common/CMakeLists.txt
@@ -28,7 +28,7 @@ if(BUILD_TESTING)
     test/msg/ArrayVal.msg
     DEPENDENCIES std_msgs)
   set(ament_cmake_copyright_FOUND TRUE)
-  ament_lint_auto_find_test_dependencies()
+  # ament_lint_auto_find_test_dependencies()
 
   ament_add_pytest_test(${PROJECT_NAME} test
     APPEND_ENV PYTHONPATH=${CMAKE_CURRENT_BINARY_DIR}
--- a/rqt_py_common/CMakeLists.txt
+++ b/rqt_py_common/CMakeLists.txt
@@ -16,7 +16,6 @@ if(BUILD_TESTING)
   find_package(rosidl_default_generators REQUIRED)
   find_package(std_msgs REQUIRED)
   find_package(ament_cmake_pytest REQUIRED)
-  find_package(ament_lint_auto REQUIRED)
 
   rosidl_generate_interfaces(${PROJECT_NAME}
     test/msg/Val.msg
--- a/rqt_py_common/CMakeLists.txt
+++ b/rqt_py_common/CMakeLists.txt
@@ -18,17 +18,10 @@ if(BUILD_TESTING)
   find_package(ament_cmake_pytest REQUIRED)
   find_package(ament_lint_auto REQUIRED)
 
-  # Disabling copyright test. The copyright used in this package does not conform to
-  # ament's copyright tests
-  set(ament_cmake_copyright_FOUND TRUE)
-  ament_lint_auto_find_test_dependencies()
-
   rosidl_generate_interfaces(${PROJECT_NAME}
     test/msg/Val.msg
     test/msg/ArrayVal.msg
     DEPENDENCIES std_msgs)
-  set(ament_cmake_copyright_FOUND TRUE)
-  # ament_lint_auto_find_test_dependencies()
 
   ament_add_pytest_test(${PROJECT_NAME} test
     APPEND_ENV PYTHONPATH=${CMAKE_CURRENT_BINARY_DIR}
--- a/rqt_py_common/CMakeLists.txt
+++ b/rqt_py_common/CMakeLists.txt
@@ -12,10 +12,6 @@ install(DIRECTORY resource
   DESTINATION share/${PROJECT_NAME}
 )
 
-install(TARGETS ${_library_name}
-  DESTINATION "${PYTHON_INSTALL_DIR}/${PROJECT_NAME}"
-)
-
 if(BUILD_TESTING)
   find_package(rosidl_default_generators REQUIRED)
   find_package(std_msgs REQUIRED)
--- a/rqt_py_common/package.xml
+++ b/rqt_py_common/package.xml
@@ -30,8 +30,6 @@
   <exec_depend>qt_gui</exec_depend>
   <exec_depend>rosidl_default_runtime</exec_depend>
 
-  <test_depend>ament_lint_auto</test_depend>
-  <test_depend>ament_lint_common</test_depend>
   <test_depend>ament_cmake_pytest</test_depend>
 
   <member_of_group>rosidl_interface_packages</member_of_group>
--- a/rqt_py_common/src/rqt_py_common/message_tree_model.py
+++ b/rqt_py_common/src/rqt_py_common/message_tree_model.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2011, Dorian Scholz, TU Darmstadt
+# Copyright (c) 2011, Dorian Scholz, TU Darmstadt  # noqa
 # All rights reserved.
 #
 # Redistribution and use in source and binary forms, with or without
@@ -38,12 +38,12 @@ from rqt_py_common.data_items import ReadonlyItem
 class MessageTreeModel(QStandardItemModel):
     _logger = logging.get_logger('MessageTreeModel')
 
-    def __init__(self, parent=None):
+    def __init__(self, parent=None):  # noqa
         # FIXME: why is this not working? should be the same as the following line...
         # super(MessageTreeModel, self).__init__(parent)
         QStandardItemModel.__init__(self, parent)
 
-    def add_message(self, message_instance, message_name='', message_type='', message_path=''):
+    def add_message(self, message_instance, message_name='', message_type='', message_path=''):  # noqa
         if message_instance is None:
             return
         self._recursive_create_items(
@@ -62,7 +62,8 @@ class MessageTreeModel(QStandardItemModel):
     def _get_data_items_for_path(self, slot_name, slot_type_name, slot_path, **kwargs):
         return (QStandardItem(slot_name), QStandardItem(slot_type_name), QStandardItem(slot_path))
 
-    def _recursive_create_items(self, parent, slot, slot_name, slot_type_name, slot_path, **kwargs):
+    def _recursive_create_items(
+            self, parent, slot, slot_name, slot_type_name, slot_path, **kwargs):
         row = []
         for item in self._get_data_items_for_path(slot_name, slot_type_name, slot_path, **kwargs):
             item._path = slot_path
@@ -76,7 +77,8 @@ class MessageTreeModel(QStandardItemModel):
                 child_slot_path = slot_path + '/' + child_slot_name
                 child_slot = getattr(slot, child_slot_name)
                 self._recursive_create_items(
-                    row[0], child_slot, child_slot_name, child_slot_type, child_slot_path, **kwargs)
+                    row[0], child_slot, child_slot_name,
+                    child_slot_type, child_slot_path, **kwargs)
 
         elif type(slot) in (list, tuple) and (len(slot) > 0):
             child_slot_type = slot_type_name[:slot_type_name.find('[')]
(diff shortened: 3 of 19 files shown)


mlautman:
https://ci.ros2.org/job/ci_linux/5490/

rqt_py_common/package.xml
```diff
@@ -33,8 +33,6 @@
   <test_depend>ament_lint_auto</test_depend>
   <test_depend>ament_lint_common</test_depend>
   <test_depend>ament_cmake_pytest</test_depend>
-  <test_depend>ament_flake8</test_depend>
-  <test_depend>ament_pep257</test_depend>
   <test_depend>python3-pytest</test_depend>
```
dirk-thomas: Same comment: `python3-pytest` doesn't seem to be used anywhere directly.

dirk-thomas (commented):
Please don't use blank `noqa`. Either ignore specific warnings in very individual cases or if you don't want make the package follow the style guide disable the linter. Adding blank noqa's all over the place isn't helpful.

mlautman:
@dirk-thomas I am only putting #noqa on existing lines of code. It would take too much time to go through each flake8 warning in the repo and fix or disable the exact warnings. All new lines of code adhere to strict flake8.

brawner:
If you want to skip flake8 tests on a file, you can place the following at the top of your file
```
# flake8: noqa 
```

dirk-thomas (changes requested):
> > Please don't use blank `noqa`.

> I am only putting #noqa on existing lines of code.

Please only suppress *specific* warnings, e.g. `# noqa: D102`.

mlautman:
@dirk-thomas  I removed the `# noqa`'s as suppressing specific warnings would take significant time.

mlautman:
https://ci.ros2.org/job/ci_linux/5536/

rqt_py_common/src/rqt_py_common/plugin_container_widget.py
```diff
@@ -100,34 +104,36 @@ def __init__(self, plugin_widget,
...
+
         @type sysmsg: str
         """
         self._logger.info('{}'.format(sysmsg))
         # self._sysmsg_widget.setPlainText(sysmsg)
         self._sysmsg_widget.append(sysmsg)
 
     def shutdown(self):
-
+        """PluginContainerWidget.shutdown."""
```
dirk-thomas: This docblock has no added value beyond the already known class name and method name. Please use `# noqa: D102` instead if there is no documentation.

Same for similar cases below.

rqt_py_common/src/rqt_py_common/topic_dict.py
```diff
@@ -32,15 +32,32 @@
...
+
+        Get the topic dictionary
+        """
         return self.topic_dict
 
     def update_topics(self, node=None):
+        """
+        update_topics.
+
+        Update the topics contained in the dictionary with new information from a node
```
dirk-thomas: This line is valuable but please remove the two line above.

rqt_py_common/src/rqt_py_common/topic_helpers.py
```diff
@@ -70,106 +77,173 @@ def get_topic_names_and_types(node=None):
 
 
 _message_class_cache = {}
-def get_message_class(message_type):
-    logger = logging.get_logger("get_message_class")
+def get_message_class(message_type):  # noqa: C901 E302
```
dirk-thomas: You don't need `E302` here. Simply at two empty lines between the cache variable and the function definition.

rqt_py_common/src/rqt_py_common/topic_helpers.py
```diff
@@ -70,106 +77,173 @@ def get_topic_names_and_types(node=None):
...
+    """
+    get_message_class: gets the message class from a string representation.
```
dirk-thomas: No need to repeat the function name:

> Get the message class from a string representation.

rqt_py_common/src/rqt_py_common/topic_dict.py
```diff
@@ -32,15 +32,32 @@
...
 
+
 class TopicDict(object):
+    """TopicDict class."""
 
     def __init__(self, node=None):
+        """
+        TopicDict(node).
+
+        Create a Topic Dict with an option node passed in
```
brawner: Add a period and use this line as the summary

rqt_py_common/src/rqt_py_common/topic_dict.py
```diff
@@ -32,15 +32,32 @@
...
+        """
         self.update_topics(node=node)
 
     def get_topics(self):
+        """
+        get_topics.
+
+        Get the topic dictionary
```
brawner: Use this as the summary line on a single line docstring.

rqt_py_common/src/rqt_py_common/topic_dict.py
brawner: Also add a period

rqt_py_common/src/rqt_py_common/topic_helpers.py
```diff
@@ -70,106 +77,173 @@ def get_topic_names_and_types(node=None):
...
+    if not _is_primative_type(type_name.lower()):
+        return get_message_class(type_name)
+
+    if type_name in ['float', 'float32', 'float64',
+                     'double', 'long double']:
         return float
 
-    elif type_name in ['string']:
-        # string constants are always stripped
+    # TODO(mlautman): char should be a string of lenght one. We do not currently suppor this
```
brawner: Spelling

brawner:
```suggestion
    # TODO(mlautman): char should be a string of length one. We do not currently support this
```

rqt_py_common/test/test_rqt_roscomm_util.py
```diff
@@ -39,18 +39,18 @@
...
-    def setUp(self):
+    def setUp(self):  # noqa
         unittest.TestCase.setUp(self)
 
-    def tearDown(self):
+    def tearDown(self):  # noqa
         unittest.TestCase.tearDown(self)
         # del self._model
 
     # TODO(mlautman): Replace this test with an apropriate ROS2 test.
```
brawner: Spelling
```suggestion
    # TODO(mlautman): Replace this test with an appropriate ROS2 test.
```

rqt_py_common/test/test_rqt_roscomm_util.py
```diff
@@ -39,18 +39,18 @@
...
     def test_iterate_packages(self):
-        """
+        """  # noqa
```
brawner: Just make this a single line docstring and add a period.

brawner (changes requested):

mlautman:
@dirk-thomas @brawner Changes made. Merge?

rqt_py_common/CMakeLists.txt
```diff
@@ -12,25 +12,17 @@ install(DIRECTORY resource
...
 if(BUILD_TESTING)
   find_package(rosidl_default_generators REQUIRED)
   find_package(std_msgs REQUIRED)
   find_package(ament_cmake_pytest REQUIRED)
   find_package(ament_lint_auto REQUIRED)
 
-  # Disabling copyright test. The copyright used in this package does not conform to
-  # ament's copyright tests
-  set(ament_cmake_copyright_FOUND TRUE)
-  ament_lint_auto_find_test_dependencies()
```
brawner, line 28: If you remove the auto tests here, you can remove the find_package line above.

rqt_py_common/src/rqt_py_common/message_tree_model.py
```diff
@@ -62,29 +62,31 @@ def _get_toplevel_items(self, index_list):
...
         for item in self._get_data_items_for_path(slot_name, slot_type_name, slot_path, **kwargs):
             item._path = slot_path
             item._user_data = kwargs.get('user_data', None)
             row.append(item)
 
         is_leaf_node = False
-        # TODO(mlautman): Work around missing _slot_types in new msg types
-        if hasattr(slot, '__slots__') and hasattr(slot, '_slot_types'):
-            for child_slot_name, child_slot_type in zip(slot.__slots__, slot._slot_types):
+        if hasattr(slot, '__slots__') and hasattr(slot, 'get_fields_and_field_types'):
```
brawner: If you check for get_fields_and_field_types you can remove the check for '\_\_slots\_\_', since that's not used here.

rqt_py_common/src/rqt_py_common/plugin_container_widget.py
```diff
@@ -46,6 +46,8 @@
 
 class PluginContainerWidget(QWidget):
     """
+    PluginContainerWidget.
```
brawner: Can you come up with a more descriptive summary line, because this will be pretty unhelpful for autodoc generation.

brawner:
```suggestion
    PluginContainerWidget visually encapsulates a plugin.
```

rqt_py_common/src/rqt_py_common/topic_dict.py
```diff
@@ -32,15 +32,20 @@
 
 from rqt_py_common.topic_helpers import get_message_class, get_topic_names_and_types
 
+
 class TopicDict(object):
+    """TopicDict class."""
```
brawner: Remove docstring, if you don't want a useful summary

brawner (commented):
Looks pretty good to me, just a couple more changes

rqt_py_common/src/rqt_py_common/topic_helpers.py
```diff
@@ -46,19 +46,26 @@ def get_topic_names_and_types(node=None):
...
     information and then destroy the node.
     """
+    # TODO(mlautman): Replace this approach with the daemon approach used by ros2cli
+    logger = logging.get_logger('get_topic_names_and_types')
     if node is not None:
-        return node.get_topic_names_and_types()
+        try:
+            return node.get_topic_names_and_types()
+        except ValueError:
+            logger.warn('method called with an invalid node!')
```
dirk-thomas: Is this exception really necessary? If a wrong `node` is passed in the method call either works or doesn't and already raises an exception in that case. Does it really make sense here to fall back to the behavior of creating a node?

mlautman:
@brawner feedback addressed

brawner (approved):

rqt_py_common/CMakeLists.txt
dirk-thomas: As well as the test dependencies in the manifest.

rqt_py_common/src/rqt_py_common/topic_helpers.py
```diff
@@ -46,19 +46,26 @@ def get_topic_names_and_types(node=None):
...
+    else:
+        logger.warn('calling get_topic_names_and_types without passing in a node is really slow!')
```
dirk-thomas: Does it make sense to even allow that? if any code wants to use it like this it also can't suppress the warning. I would suggest to remove this option.

rqt_py_common/src/rqt_py_common/topic_helpers.py
```diff
@@ -70,106 +77,175 @@ def get_topic_names_and_types(node=None):
...
+    topic_names_and_types = get_topic_names_and_types(node=node)
+    return _get_field_type(topic_names_and_types, target)
+
+
+def _get_field_type(topic_names_and_types, target):  # noqa: C901
+    """Testable helper function for get_field_type."""
+    logger = logging.get_logger('topic_helpers._get_field_type')
+    for name, types in topic_names_and_types:
+        # If the topic is a substring of the target param, then we have found a match
+        if target.find(name) >= 0:
```
dirk-thomas: I don't understand why `target.find(name) > 0` would be a match? In the following code you expect `target == name` or use it as `target[len(name):]`.

rqt_py_common/src/rqt_py_common/topic_helpers.py
```diff
@@ -70,106 +77,175 @@ def get_topic_names_and_types(node=None):
...
+            if target == name:
+                # If there is more than one type of topic on target
+                if len(types) > 1:
+                    logger.warn(
+                        'Ambiguous request. Multiple topic types found on: {}'.format(target))
+                    return None, False
+                # If the types array is empty then something weird has happend
+                if len(types) == 0:
+                    logger.warn(
+                        'Ambiguous request. No topic types found on: {}'.format(target))
```
dirk-thomas: `0` doesn't sounds "Ambiguous".

rqt_py_common/src/rqt_py_common/topic_helpers.py
mlautman: This is temporary, I already have notes for why this is a bad idea and a todo to replace it. I'm not inclined to fix this here as this PR is already really really big and a better solution is already available in @brawner's changes https://github.com/ros-visualization/rqt/pull/161. I removed the fall-back logic

rqt_py_common/src/rqt_py_common/topic_helpers.py
mlautman: At that point the entire method is rather useless I looked through the rqt packages and I think we can remove this method entirely

rqt_py_common/src/rqt_py_common/topic_helpers.py
mlautman: I changed this line of logic to `target.startswith(name)`

mlautman:
@dirk-thomas comments addressed.

rqt_py_common/src/rqt_py_common/topic_helpers.py
dirk-thomas: The updated condition still seems to match unrelated ones. E.g. `target` being `/foobar/baz` and `name` being `/foo`.

mlautman: Good point. New commit covers that corner case and tests for it as well

mlautman:
@dirk-thomas This pr is blocking both me and Stephen. The sooner we can get this merged the better

mlautman:
This was blocking both me and @brawner. After many rounds of reviews, I am merging it with only one approval

PR #156, opened 2018-11-09
Adjusting message_tree_model for ros2

brawner:
Here is one solution to addressing the missing _type field in generated python message files. This PR lets me get rqt_publisher to work. It checks to make sure the assumed property name exists and if it doesn't it will use the slot name, which currently start with underscores.

### DIFF
--- a/rqt_py_common/src/rqt_py_common/message_tree_model.py
+++ b/rqt_py_common/src/rqt_py_common/message_tree_model.py
@@ -62,6 +62,18 @@ class MessageTreeModel(QStandardItemModel):
     def _get_data_items_for_path(self, slot_name, slot_type_name, slot_path, **kwargs):
         return (QStandardItem(slot_name), QStandardItem(slot_type_name), QStandardItem(slot_path))
 
+    def _get_property_name(self, slot_name, message_instance):
+        # slots start with an underscore
+        property_name = slot_name[1:]
+
+        # In ROS2 the user friendly name is a property, check to make sure that field exists
+        if hasattr(message_instance.__class, property_name) and \
+                isinstance(getattr(message_instance.__class__, property_name), property):
+            return property_name
+
+        # Otherwise, we'll just use the slot name
+        return slot_name
+
     def _recursive_create_items(self, parent, slot, slot_name, slot_type_name, slot_path, **kwargs):
         row = []
         for item in self._get_data_items_for_path(slot_name, slot_type_name, slot_path, **kwargs):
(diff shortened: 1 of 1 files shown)


rqt_py_common/src/rqt_py_common/message_tree_model.py
```diff
@@ -70,13 +82,14 @@ def _recursive_create_items(self, parent, slot, slot_name, slot_type_name, slot_
             row.append(item)
 
         is_leaf_node = False
-        # TODO(mlautman): Work around missing _slot_types in new msg types
```
mlautman: I actually fixed this with https://github.com/ros2/rosidl_python/pull/19

mlautman, line 73: I actually fixed this with https://github.com/ros2/rosidl_python/pull/19 

I thought I had told you about that but maybe not... We need a good system for communicating things like this. Also we should start sharing the repos file for testing so that we are both building/testing against the same ros2 env

rqt_py_common/src/rqt_py_common/message_tree_model.py
```diff
@@ -62,6 +62,18 @@ def _get_toplevel_items(self, index_list):
     def _get_data_items_for_path(self, slot_name, slot_type_name, slot_path, **kwargs):
         return (QStandardItem(slot_name), QStandardItem(slot_type_name), QStandardItem(slot_path))
 
+    def _get_property_name(self, slot_name, message_instance):
```
mlautman, line 65: no need to pass in self. Could be a static function

rqt_py_common/src/rqt_py_common/message_tree_model.py
```diff
@@ -62,6 +62,18 @@ def _get_toplevel_items(self, index_list):
...
+        # slots start with an underscore
+        property_name = slot_name[1:]
+
+        # In ROS2 the user friendly name is a property, check to make sure that field exists
+        if hasattr(message_instance.__class, property_name) and \
```
mlautman, line 70: \_\_class\_\_

rqt_py_common/src/rqt_py_common/message_tree_model.py
```diff
@@ -62,6 +62,18 @@ def _get_toplevel_items(self, index_list):
     def _get_data_items_for_path(self, slot_name, slot_type_name, slot_path, **kwargs):
         return (QStandardItem(slot_name), QStandardItem(slot_type_name), QStandardItem(slot_path))
 
+    def _get_property_name(self, slot_name, message_instance):
+        # slots start with an underscore
+        property_name = slot_name[1:]
```
mlautman, line 67: You shouldn't hardcode this without first checking. First check the length >1 and that the first character is an underscore.

rqt_py_common/src/rqt_py_common/message_tree_model.py
```diff
@@ -70,13 +82,14 @@ def _recursive_create_items(self, parent, slot, slot_name, slot_type_name, slot_
...
         is_leaf_node = False
-        # TODO(mlautman): Work around missing _slot_types in new msg types
-        if hasattr(slot, '__slots__') and hasattr(slot, '_slot_types'):
-            for child_slot_name, child_slot_type in zip(slot.__slots__, slot._slot_types):
-                child_slot_path = slot_path + '/' + child_slot_name
-                child_slot = getattr(slot, child_slot_name)
+        if hasattr(slot, '__slots__'):
+            for child_slot_name in slot.__slots__:
+                child_property_name = self._get_property_name(child_slot_name, slot)
+                child_slot_type = type(getattr(slot, child_property_name)).__name__
```
mlautman, line 88: This returns list when what we actually want is `rqt_py_common/Val[5]` I already fixed this and implemented the solution in a pr to rosidl_python

mlautman (changes requested):

brawner:
Closing in favor of https://github.com/ros2/rosidl_python/pull/19

PR #158, opened 2018-11-10
Get field types

mlautman:
This pr includes #155 

New commit: 3b61b6e

This implements the get_field_type method.
This also causes the rqt_py_common package to depend on ros2cli packages to make use of the daemon node for getting topics and message types. Ideally this functionality will get split out of ros2cli in the future so that the RQT package no longer depends on a cli package. 

https://ci.ros2.org/job/ci_linux/5512/

### DIFF
--- a/rqt_py_common/src/rqt_py_common/topic_helpers.py
+++ b/rqt_py_common/src/rqt_py_common/topic_helpers.py
@@ -46,8 +46,15 @@ def get_topic_names_and_types(node=None):
     If node is None, then this method will create a node, use it to get topic
     information and then destroy the node.
     """
+    # TODO(mlautman): Replace this approach with the daemon approach used by ros2cli
+    logger = logging.get_logger('get_topic_names_and_types')
     if node is not None:
-        return node.get_topic_names_and_types()
+        try:
+            return node.get_topic_names_and_types()
+        except ValueError:
+            logger.warn('method called with an invalid node!')
+    else:
+        logger.warn('calling get_topic_names_and_types without passing in a node is really slow!')
 
     import rclpy
     shutdown_rclpy = False
--- a/rqt_py_common/src/rqt_py_common/topic_helpers.py
+++ b/rqt_py_common/src/rqt_py_common/topic_helpers.py
@@ -55,7 +55,7 @@ def get_topic_names_and_types(node=None):
         shutdown_rclpy = True
         rclpy.init()
 
-    node = rclpy.create_node("TopicHelpers__get_topic_names_and_types")
+    node = rclpy.create_node('TopicHelpers__get_topic_names_and_types')
 
     # Give the node time to learn about the graph
     rclpy.spin_once(node, timeout_sec=0.5)
@@ -69,22 +69,28 @@ def get_topic_names_and_types(node=None):
     return topic_list
 
 
-_message_class_cache = {}
+_message_class_cache = {}  # noqa
 def get_message_class(message_type):
-    logger = logging.get_logger("get_message_class")
+    """
+    get_message_class: gets the message class from a string representation.
+
+    @param message_type: the type of message in the form `msg_pkg/Message`
+    @type message_type: str
+    """
+    logger = logging.get_logger('get_message_class')
     if message_type in _message_class_cache:
         return _message_class_cache[message_type]
 
-    message_info = message_type.split("/")
+    message_info = message_type.split('/')
     if len(message_info) == 2:
         package = message_info[0]
         base_type = message_info[1]
     elif len(message_info) == 1:
-        package = "std_msgs"
+        package = 'std_msgs'
         base_type = message_info[0]
     else:
         logger.error(
-            "Malformed message_type passed into get_message_class: {}".format(
+            'Malformed message_type passed into get_message_class: {}'.format(
                 message_type))
         return None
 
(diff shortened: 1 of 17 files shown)


rqt_py_common/src/rqt_py_common/topic_helpers.py
```diff
@@ -46,130 +50,259 @@ def get_topic_names_and_types(node=None):
...
-    #     return None, False
-
-    # slot_path = topic_name[len(real_topic_name):]
-    # return get_slot_type(message_class, slot_path)
-    pass
+    topic_names_and_types = get_topic_names_and_types(node=node)
+    return _get_field_type(topic_names_and_types, target)
+
+
+
```
brawner: Should be two lines of space between functions

rqt_py_common/src/rqt_py_common/topic_helpers.py
```diff
@@ -46,130 +50,259 @@ def get_topic_names_and_types(node=None):
...
+def _get_field_type(topic_names_and_types, target):  # noqa
+    """
+    Testable helper function for get_field_type.
+    """
+    logger = logging.get_logger('topic_helpers._get_field_type')
+    for name, types in topic_names_and_types:
```
brawner, line 204: It would probably be faster turn this into a dictionary and just check if the dictionary contains the key. This is already a list of (name, types) tuples, so a making a dict is straight forward.

rqt_py_common/src/rqt_py_common/topic_helpers.py
```diff
@@ -46,130 +50,259 @@ def get_topic_names_and_types(node=None):
...
+        try:
+            return node.get_topic_names_and_types()
+        except ValueError:
+            logger.warn('method called with an invalid node!')
 
-    node.destroy_node()
-    if shutdown_rclpy:
-        rclpy.shutdown()
+    else:
+        with NodeStrategy([]) as node:
```
brawner: This throws an exception of rclpy is already initialized, but a node has not been passed in as an argument.

rqt_py_common/src/rqt_py_common/topic_helpers.py
```diff
@@ -189,6 +322,6 @@ def get_slot_type(message_class, slot_path):
     for field_name in fields:
         slot_class_name = message_class._slot_types[message_class.__slots__.index(field_name)]
         is_array = slot_class_name.find('[') >= 0
-        message_class = get_message_class(slot_class_name[:slot_class_name.find('[')])
+        message_class = get_type_class(slot_class_name[:slot_class_name.find('[')])
```
brawner, line 267: This creates an off-by-one error if find returns -1.

rqt_py_common/src/rqt_py_common/topic_helpers.py
```diff
@@ -46,130 +50,259 @@ def get_topic_names_and_types(node=None):
...
-    elif type_name in ["bool"]:
+    if type_name in ['bool', 'boolean']:
         return bool
 
-    else:
-        return None
+    return None
 
 
-def get_field_types(topic_name):
```
brawner, line 142: Why remove this? What if someone wants all the types?

brawner (changes requested):

rqt_py_common/src/rqt_py_common/topic_helpers.py
brawner: Nevermind, I get what this is trying to do.

rqt_py_common/src/rqt_py_common/topic_helpers.py
```diff
@@ -46,130 +50,259 @@ def get_topic_names_and_types(node=None):
...
+
+                # Iterate through the message types on the given topic and see if any match the
+                # path that was provided
+                for msg_type_str in types:
+                    try:
+                        msg_class = get_message_class(msg_type_str)
+
+                        # If the message type is a simple type eg: float, then it can't be an array
+                        if _is_primative_type(msg_type_str):
+                            return msg_class, False
```
brawner, line 239: ```get_field_type``` is supposed to return the Python type of a field, not the full message class. For example, /chatter/_data resolves to std_msgs.msg._string.String

rqt_py_common/src/rqt_py_common/topic_helpers.py
brawner: You're also calling .find twice here. You can just store the result of the first one to a variable

rqt_py_common/src/rqt_py_common/topic_helpers.py
mlautman: The original implementation was get_field_type. https://github.com/ros-visualization/rqt/blob/ffcab927b4c1a1b359e269012057b3048a9af685/rqt_py_common/src/rqt_py_common/topic_helpers.py#L50

mlautman:
Closing since the changes got mixed into #155. That PR will now contain both some cleanup and this new feature

rqt_py_common/src/rqt_py_common/topic_helpers.py
mlautman: Good catch. Addressed in a separate PR

rqt_py_common/src/rqt_py_common/topic_helpers.py
mlautman: Addressed in a separate PR

rqt_py_common/src/rqt_py_common/topic_helpers.py
mlautman: Addressed in a separate PR

rqt_py_common/src/rqt_py_common/topic_helpers.py
mlautman: Addressed in #155

PR #161, opened 2018-11-15
Subclassing plugin_context to pass around the rqt node

brawner:
This adds a derived class Ros2PluginContext, to pass around a ros2 node so that plugins have the ability to communicate with the node graph.

### DIFF
--- a/rqt_gui/src/rqt_gui/ros2_plugin_context.py
+++ b/rqt_gui/src/rqt_gui/ros2_plugin_context.py
@@ -35,9 +35,10 @@ class Ros2PluginContext(PluginContext):
     """
     Provides information to the plugin and exposes methods to interact with the framework.
 
-    PluginContext relays all methods to the corresponding `PluginHandler`.
+    Ros2PluginContext exposes the shared rclpy Node to rqt_gui plugins and relays all methods to the
+    correspending `PluginHandler`
     """
 
     def __init__(self, handler, node):
         super(Ros2PluginContext, self).__init__(handler)
-        self._node = node
+        self.node = node
--- a/rqt_gui/src/rqt_gui/ros2_plugin_context.py
+++ b/rqt_gui/src/rqt_gui/ros2_plugin_context.py
@@ -1,0 +1,43 @@
+# Copyright (c) 2018 PickNik Consulting
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions
+# are met:
+#
+#   * Redistributions of source code must retain the above copyright
+#     notice, this list of conditions and the following disclaimer.
+#   * Redistributions in binary form must reproduce the above
+#     copyright notice, this list of conditions and the following
+#     disclaimer in the documentation and/or other materials provided
+#     with the distribution.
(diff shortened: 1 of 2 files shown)


mlautman (approved):

mlautman:
Looks good to me

rqt_gui/src/rqt_gui/ros2_plugin_context.py
```diff
@@ -0,0 +1,142 @@
...
+
+
+class Ros2PluginContext(PluginContext):
+    """
+    Provides information to the plugin and exposes methods to interact with the framework.
+
+    PluginContext relays all methods to the corresponding `PluginHandler`.
+    """
+
+    def __init__(self, handler, node):
```
mlautman, line 42: See my comments on https://github.com/ros-visualization/rqt_top/pull/6 about these helper functions. It seems like it would be better to simply pass out a copy of the node... Any reason why we wouldn't want to do that. 

..I realize that I previously requested the get_topic_names_and_types method here but maybe that was a bad idea

brawner: The concern is just that someone writes a plugin and calls a mutable method on node without realizing the implications of doing so. That will impact the other plugins running on their system.

There is no way to prevent this from happening 100%, but a wrapper would heavily discourage calling these mutable functions. I did leave in the create_xyz and destroy_xyz methods because it's fair to expect the plugin to keep track of creation and destruction of the publishers/subscribers/clients/services that they need.

Another option would be to create a copy and overwrite __class__ of the copied node with a subclass that unimplements the methods that should not be available to plugins.

If they want a node that they can directly access, the plugin can always create one themselves.

brawner:
Fixed the imports

rqt_gui_py/src/rqt_gui_py/ros_py_plugin_provider.py
```diff
@@ -46,12 +47,17 @@ def __init__(self):
         self._node_initialized = False
         self._node = None
 
+    def __del__(self):
+        self._destroy_node()
```
dirk-thomas, line 51: The node needs to be destroyed earlier since this destructor is not guaranteed to be called.

rqt_gui/src/rqt_gui/ros2_plugin_context.py
dirk-thomas: I am not convinced this is necessary. The big disadvantage is that this needs to be updated whenever a new method is added. E.g. currently the getter for the clock is missing from that interface. On the other hand it can't prevent users from accessing undesired methods anyway since they can just get `_node`.

Therefore I suggest to just add `node` to the context. The same needs to happen for the C++ equivalent.

brawner: Sounds good. Will revert it back

rqt_gui_py/src/rqt_gui_py/ros_py_plugin_provider.py
brawner: Adding it to the derived method shutdown() instead

mlautman:
Merging with one approval since Dirk is super busy for the next few days

PR #162, opened 2018-11-16
Rqt py common message helpers

mlautman:
includes #155 
New changes: 467f10c5072bc278c71f88008899e6a5c675b480

### DIFF
--- a/rqt_py_common/src/rqt_py_common/message_helpers.py
+++ b/rqt_py_common/src/rqt_py_common/message_helpers.py
@@ -40,8 +40,13 @@ MSG_MODE = 'msg'
 SRV_MODE = 'srv'
 ACTION_MODE = 'action'
 
-# Taken from https://github.com/ros2/ros2cli/blob/bouncy/ros2msg/ros2msg/api/__init__.py
+
 def get_all_service_types():
+    """
+    Uses the ament index to iterate through packages and calls get_service_types on each one.
+
+    :returns: a dictionary of the form {'package_name', ['srv1', 'srv2', ...]}
+    """
     all_service_types = {}
     for package_name in get_resources('rosidl_interfaces').keys():
         service_types = get_service_types(package_name)
@@ -51,6 +56,12 @@ def get_all_service_types():
 
 
 def get_service_types(package_name):
+    """
+    Uses the ament index gind all services avialable in the package.
+
+    :param package_name: a string eg 'std_srvs'
+    :returns: a dictionary of the form {'package_name', ['srv1', 'srv2', ...]}
+    """
     if not has_resource('packages', package_name):
         raise LookupError('Unknown package name')
     try:
@@ -68,6 +79,11 @@ def get_service_types(package_name):
 
 
 def get_all_message_types():
+    """
+    Uses the ament index to iterate through packages and calls get_message_types on each one.
+
+    :returns: a dictionary of the form {'package_name', ['msg1', 'msg2', ...]}
+    """
     all_message_types = {}
     for package_name in get_resources('rosidl_interfaces').keys():
         message_types = get_message_types(package_name)
@@ -77,6 +93,12 @@ def get_all_message_types():
 
 
 def get_message_types(package_name):
+    """
+    Uses the ament index gind all messages avialable in the package.
+
+    :param package_name: a string eg 'std_msgs'
+    :returns: a dictionary of the form {'std_msgs', ['Bool', 'String', ...]}
+    """
     if not has_resource('packages', package_name):
         raise LookupError('Unknown package name')
     try:
@@ -93,46 +115,52 @@ def get_message_types(package_name):
     return all_msgs
 
 
-def _get_message_service_class_helper(message_type, mode, logger):
+def _get_message_service_class_helper(message_type, mode, logger=None):  # noqa: C901
+    """
+    A helper function for common logic to be used by get_message_class and get_service_class.
+
+    :param message_type: name of the message or service class in the form 'package_name/msg_name'
+    :type message_type: str
+    :param mode: one of MSG_MODE or SRV_MODE
+    :type mode: str
+    :param logger: The logger to be used for warnings and info
+    :type logger: either rclpy.impl.rcutils_logger.RcutilsLogger or None
+
+    :returns: The message or service class or None
+    """
+    if logger is None:
+        logger = logging.get_logger('_get_message_service_class_helper')
+
     if not (mode == MSG_MODE or mode == SRV_MODE):
         logger.warn('invalid mode {}'.format(mode))
         return None
 
     message_info = message_type.split('/')
-    if len(message_info) == 2:
-        package = message_info[0]
-        base_type = message_info[1]
-    elif len(message_info) == 1:
-        package = 'std_' + mode + 's'
-        base_type = message_info[0]
-    else:
-        logger.error(
-            'Malformed message_type: {}'.format(message_type))
+    if len(message_info) != 2:
+        logger.error('Malformed message_type: {}'.format(message_type))
         return None
 
+    package = message_info[0]
+    base_type = message_info[1]
+
     _, resource_path = get_resource('rosidl_interfaces', package)
-    python_pkg = class_val = None
+    python_pkg = None
+    class_val = None
+
     try:
         # import the package
         python_pkg = __import__('%s.%s' % (package, mode))
     except ImportError:
-        logger.error('Failed to import class: {} as {}.{}'.format(message_type, package, mode))
+        logger.info('Failed to import class: {} as {}.{}'.format(message_type, package, mode))
+        return None
 
-    if python_pkg:
-        try:
-            class_val = getattr(getattr(python_pkg, mode), base_type)
-        except AttributeError:
-            if len(base_type):
-                # Try the same thing but with the first letter capitalized
-                base_type = ''.join([base_type[0].upper(), base_type[1:]])
+    try:
+        class_val = getattr(getattr(python_pkg, mode), base_type)
+        return class_val
 
-        if not class_val:
-            try:
-                class_val = getattr(getattr(python_pkg, mode), base_type)
-            except AttributeError:
-                logger.error('Failed to load class: {}'.format(message_type))
-
-    return class_val
+    except AttributeError:
+        logger.info('Failed to load class: {}'.format(message_type))
+        return None
 
 
 _srv_class_cache = {}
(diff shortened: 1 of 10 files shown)


rqt_py_common/src/rqt_py_common/message_helpers.py
```diff
@@ -0,0 +1,200 @@
...
+from ament_index_python import has_resource
+
+from rclpy import logging
+
+MSG_MODE = 'msg'
+SRV_MODE = 'srv'
+ACTION_MODE = 'action'
+
+# Taken from https://github.com/ros2/ros2cli/blob/bouncy/ros2msg/ros2msg/api/__init__.py
+def get_all_service_types():
```
brawner, line 45: Flake8 requires two lines above a function

rqt_py_common/src/rqt_py_common/message_helpers.py
```diff
@@ -0,0 +1,200 @@
...
+    class_val = _get_message_service_class_helper(message_type, MSG_MODE, logger)
+
+    if class_val is not None:
+        _message_class_cache[message_type] = class_val
+
+    return class_val
+
+
+def get_message_text_from_class(msg_class):
+    msg_slot_dict = msg_class.get_slot_types_dict()
```
brawner: Update based on upstream PR

rqt_py_common/src/rqt_py_common/message_helpers.py
```diff
@@ -0,0 +1,200 @@
...
+    return ''.join(
+        ['{0:{width}}{1}\n'.format(slot_type, slot_name, width=30) for
+            slot_name, slot_type in msg_slot_dict.items()])
+
+
+def get_service_text_from_class(srv_class):
+    srv_slot_dict_req = srv_class.Request.get_slot_types_dict()
+    srv_slot_dict_res = srv_class.Response.get_slot_types_dict()
```
brawner: Update these two lines based on upstream PR

rqt_py_common/src/rqt_py_common/message_helpers.py
```diff
@@ -0,0 +1,200 @@
...
+from rclpy import logging
+
+MSG_MODE = 'msg'
+SRV_MODE = 'srv'
+ACTION_MODE = 'action'
+
+# Taken from https://github.com/ros2/ros2cli/blob/bouncy/ros2msg/ros2msg/api/__init__.py
+def get_all_service_types():
+    all_service_types = {}
+    for package_name in get_resources('rosidl_interfaces').keys():
```
brawner, line 52: It's probably worth it to call this function only once between this function and get_service_types

rqt_py_common/src/rqt_py_common/message_helpers.py
```diff
@@ -0,0 +1,200 @@
...
+        service_types = get_service_types(package_name)
+        if service_types:
+            all_service_types[package_name] = service_types
+    return all_service_types
+
+
+def get_service_types(package_name):
```
brawner, line 59: Regarding the above comment, maybe add a 'resources' parameter with a default None value

rqt_py_common/src/rqt_py_common/message_helpers.py
```diff
@@ -0,0 +1,200 @@
...
+    all_srvs = [
+        n[4:-4] for n in interface_names if n.startswith('srv/') and n.endswith('.srv')]
+    all_srvs.extend(
+        [n[:-4] for n in interface_names if not n.startswith('srv/') and n.endswith('.srv')])
+    return all_srvs
+
+
+def get_all_message_types():
+    all_message_types = {}
+    for package_name in get_resources('rosidl_interfaces').keys():
```
brawner, line 87: Same comment about these functions (could reuse call to get_resources)

rqt_py_common/src/rqt_py_common/message_helpers.py
```diff
@@ -0,0 +1,200 @@
...
+        except AttributeError:
+            if len(base_type):
+                # Try the same thing but with the first letter capitalized
+                base_type = ''.join([base_type[0].upper(), base_type[1:]])
+
+        if not class_val:
+            try:
+                class_val = getattr(getattr(python_pkg, mode), base_type)
+            except AttributeError:
+                logger.error('Failed to load class: {}'.format(message_type))
```
brawner: Include exception in error message. It won't be clear why this failed in the error message otherwise

brawner (dismissed):

rqt_py_common/src/rqt_py_common/message_helpers.py
mlautman: get_all_service_types calls `get_resources`, get_service_types calls `get_resource`

rqt_py_common/src/rqt_py_common/message_helpers.py
mlautman: Seems like premature optimization

rqt_py_common/src/rqt_py_common/message_helpers.py
mlautman: I'm not sure I agree. I think that if you call get_message_class("std_msgs/not_a_msg") it should tell you that it failed to load and return `None`. The full traceback would be spammy

rqt_py_common/src/rqt_py_common/message_helpers.py
brawner: Right, but you call get_service_types in this function.

rqt_py_common/src/rqt_py_common/message_helpers.py
brawner: Considering get_resources scales  linearly with the number of packages installed and you are following it up with calls to has_resource and get_resource, it seems like optimization should be considered. These functions require OS calls.

rqt_py_common/src/rqt_py_common/message_helpers.py
```diff
@@ -0,0 +1,234 @@
...
+    for package_name in get_resources('rosidl_interfaces').keys():
+        message_types = get_message_types(package_name)
+        if message_types:
+            all_message_types[package_name] = message_types
+    return all_message_types
+
+
+def get_message_types(package_name):
+    """
+    Uses the ament index gind all messages avialable in the package.
```
brawner: Spelling

rqt_py_common/src/rqt_py_common/message_helpers.py
```diff
@@ -0,0 +1,234 @@
...
+    package = message_info[0]
+    base_type = message_info[1]
+
+    _, resource_path = get_resource('rosidl_interfaces', package)
+    python_pkg = None
+    class_val = None
+
+    try:
+        # import the package
+        python_pkg = __import__('%s.%s' % (package, mode))
```
brawner: Use importlib

rqt_py_common/test/test_rqt_message_helpers.py
```diff
@@ -0,0 +1,84 @@
...
+# LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+# ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+
+# Author: Michael Lautman
+
+import unittest
+
+
+class TestMessageHelpers(unittest.TestCase):  # noqa: D101
```
brawner, line 40: You can just add:
```
# flake8: noqa
```
So you don't have to ignore all warnings

https://github.com/ros-visualization/qt_gui_core/blob/0c2aaaaa4340f2a810abd91f6f328236c6beb1f9/qt_dotgraph/test/dot_to_qt_test.py#L35

rqt_py_common/test/test_rqt_topic_helpers.py
```diff
@@ -37,25 +37,12 @@
 import unittest
 
 
-class TestTopicHelpers(unittest.TestCase):
+class TestTopicHelpers(unittest.TestCase):  # noqa: D101
```
brawner, line 40: Same comment about ```# flake8: noqa```

rqt_py_common/src/rqt_py_common/topic_helpers.py
```diff
@@ -30,66 +30,12 @@
...
-                logger.error('Failed to get message class: {}'.format(message_type))
-
-    if class_val:
-        _message_class_cache[message_type] = class_val
-
-    return class_val
-
-
-def _is_primative_type(type_str):
+def is_primative_type(type_str):
```
brawner: is_primitive_type
```suggestion
def is_primitive_type(type_str):
```

rqt_py_common/src/rqt_py_common/topic_helpers.py
```diff
@@ -225,9 +169,15 @@ def get_slot_type(message_class, slot_path):
...
             is_array = True
-            message_class = get_type_class(slot_class_name[:array_index])
+            if is_primative_type(slot_class_name[:array_index]):
+                message_class = get_type_class(slot_class_name[:array_index])
+            else:
+                message_class = get_message_class(slot_class_name[:array_index])
         else:
             is_array = False
-            message_class = get_type_class(slot_class_name)
+            if is_primative_type(slot_class_name):
```
brawner: is_primitive_type
```suggestion
            if is_primitive_type(slot_class_name):
```

rqt_py_common/src/rqt_py_common/topic_helpers.py
```diff
@@ -225,9 +169,15 @@ def get_slot_type(message_class, slot_path):
         array_index = slot_class_name.find('[')
         if array_index >= 0:
             is_array = True
-            message_class = get_type_class(slot_class_name[:array_index])
+            if is_primative_type(slot_class_name[:array_index]):
```
brawner: is_primitive_type
```suggestion
            if is_primitive_type(slot_class_name[:array_index]):
```

rqt_py_common/src/rqt_py_common/message_helpers.py
mlautman: I don't understand. get_resources returns a dictionary mapping package names to file paths. I pass in the list of package names to get_resource which returns a tuple with the package resources first. How else would I get a list of all available packages so that I can call get_resource to get the individual msg resources?

Also, I'm inclined to leave it as we have yet to experience any performance issues.

rqt_py_common/src/rqt_py_common/message_helpers.py
mlautman: Linear isn't a big deal in this case. Even if we have a thousand packages, RQT only needs to run at human speed. I removed the has_resource call and the try except block since I was already catching exceptions thrown by get_resource.

rqt_py_common/test/test_rqt_message_helpers.py
mlautman: Dirk wasn't a fan of disabling all warnings. This works fine for me

PR #164, opened 2018-11-20
Port rqt gui cpp content (porting roscpp_plugin_provider)

mlautman:
I am hoping to get some early feedback on bba4171 if you have a chance


dirk-thomas:
> I am hoping to get some early feedback on bba4171 if you have a chance

In Python we decided to provide a node handle through the context. Why should each C++ plugin *be* a node?

Also from the patch alone it isn't clear yet how the executor is being added.

Please use a separate commit to rename files (without changing their content).

PR #163, opened 2018-11-20
Porting rqt_gui_cpp Cmake

mlautman:
This pr ports rqt_gui_cpp to be built with Colcon 

I intentionally separated this from porting the plugin providers so that we can get a jump start on the review process. Also, seeing as this is my first cpp port, I wanted to get a second pair of eyes on just the cmake changes

PR #167, opened 2018-12-06
Spinning node in separate QThread

brawner:
The node that was added to the plugin_context was not spinning previously so Python subscribers would not work correctly. This now adds a QThread to process the events during spin.

### DIFF
--- a/rqt_gui_py/src/rqt_gui_py/rclpy_spinner.py
+++ b/rqt_gui_py/src/rqt_gui_py/rclpy_spinner.py
@@ -1,0 +1,38 @@
+# Copyright 2018, PickNik Consulting
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from python_qt_binding.QtCore import QThread, qWarning, qDebug
+import rclpy
+from rclpy.executors import MultiThreadedExecutor
+
+
+class RclpySpinner(QThread):
+    def __init__(self, node):
+        super().__init__()
+        self._node = node
+        self._abort = False
+
+    def run(self):
+        qDebug('Start called on RclpySpinner, spinning ros2 node')
+        executor = MultiThreadedExecutor()
+        executor.add_node(self._node)
+        while rclpy.ok() and not self._abort:
+            executor.spin_once(timeout_sec=1.0)
+        if not self._abort:
+            qWarning('rclpy.shutdown() was before quit was called on this QThread')
+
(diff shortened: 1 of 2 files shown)


mlautman (commented):
+1 after demoing with another rqt plugin and passing ci

rqt_gui_py/src/rqt_gui_py/ros_py_plugin_provider.py
```diff
@@ -65,8 +68,11 @@ def _init_node(self):
         if not self._node_initialized:
             name = 'rqt_gui_py_node_%d' % os.getpid()
             qDebug('RosPyPluginProvider._init_node() initialize ROS node "%s"' % name)
+
             rclpy.init()
             self._node = rclpy.create_node(name)
```
mlautman, line 73: A possible improvement: Add a ReentrantCallbackGroup to the context so that multiple plugins can have callbacks serviced simultaneously.

mlautman (approved):

rqt_gui_py/src/rqt_gui_py/rclpy_spinner.py
```diff
@@ -0,0 +1,38 @@
...
+        self._abort = False
+
+    def run(self):
+        qDebug('Start called on RclpySpinner, spinning ros2 node')
+        executor = MultiThreadedExecutor()
+        executor.add_node(self._node)
+        while rclpy.ok() and not self._abort:
+            executor.spin_once(timeout_sec=1.0)
+        if not self._abort:
+            qWarning('rclpy.shutdown() was before quit was called on this QThread')
```
dirk-thomas, line 33: This doesn't sounds like a sentence - maybe a word missing?

rqt_gui_py/src/rqt_gui_py/ros_py_plugin_provider.py
```diff
@@ -65,8 +68,11 @@ def _init_node(self):
...
+            self.spinner = RclpySpinner(self._node)
```
dirk-thomas, line 74: This line should probably go into the constructor to ensure `spinner` is always set. Otherwise `shutdown` can't assume it exists.

rqt_gui_py/src/rqt_gui_py/ros_py_plugin_provider.py
dirk-thomas: A `ReentrantCallbackGroup` implies that the very same callback can be called multiple times (not that "multiple plugins can have callbacks serviced simultaneously"). Usually that is undesired and most callback function won't be written in a way to support that.

rqt_gui_py/src/rqt_gui_py/ros_py_plugin_provider.py
brawner: This was addressed in PR #163

rqt_gui_py/src/rqt_gui_py/rclpy_spinner.py
brawner: Addressed in #171

rqt_gui_py/src/rqt_gui_py/ros_py_plugin_provider.py
brawner: That's good to know Dirk. We'll close the issue.

Issue #166, opened 2018-12-04
ROS2 port: migrate to pure python package


brawner:
In order to open an rqt_gui window using ```ros2 run rqt_gui rqt_gui``` it needs to be compiled as an application (Windows can't execute scripts without a provided interpreter and ignores the shebang). It was recommended to create a pure python package for rqt_gui instead of an ament_cmake package.

dirk-thomas:
Addressed in #170.

PR #169, opened 2018-12-07
Moving rqt to pure python

brawner:
To enable bundling of python scripts (rqt, rqt_gui) into executables on Windows and to significantly speed up build time, this PR converts rqt, rqt_gui and rqt_gui_py to ament_python packages.

This addresses issue https://github.com/ros-visualization/rqt/issues/166


rqt/setup.py
```diff
@@ -0,0 +1,29 @@
+from setuptools import setup
+
+package_name = 'rqt'
+setup(
+    name=package_name,
+    version='0.5.0',
+    install_requires=['rqt_gui', 'rqt_gui_cpp', 'rqt_gui_py', 'rqt_py_common'],
+    data_files=[
+        ('share/ament_index/resource_index/packages',
+            ['resource/' + package_name]),
```
mlautman: is this actually needed? launch is a pure python package and it doesn't do this empty file thing

rqt_gui/setup.py
```diff
@@ -0,0 +1,38 @@
...
+
+package_name = 'rqt_gui'
+setup(
+    name=package_name,
+    version='0.5.0',
+    package_dir={'': 'src'},
+    packages=['rqt_gui'],
+    data_files=[
+        ('share/ament_index/resource_index/packages',
+            ['resource/' + package_name]),
```
mlautman: same as above

rqt_gui/setup.py
```diff
@@ -0,0 +1,38 @@
+from setuptools import setup
+
+package_name = 'rqt_gui'
+setup(
+    name=package_name,
+    version='0.5.0',
+    package_dir={'': 'src'},
+    packages=['rqt_gui'],
+    data_files=[
+        ('share/ament_index/resource_index/packages',
```
mlautman: Why does this work on Windows?? Shouldn't it be `share\ament_index\...`?

rqt/setup.py
brawner: You're right, this is no longer needed it seems.

rqt_gui/setup.py
brawner: Doesn't look like this is needed.

rqt_gui/setup.py
brawner: setuptools takes care of the cross-platform bit of this code.

dirk-thomas:
The Python packages need to register themselves at the resource index as well as install their manifest file: see https://github.com/ros2/demos/blob/f2a1d9f5fe694a11fb42f3c036dacd130141ac0e/demo_nodes_py/setup.py#L10-L14 for an example.

Also since the patch removed all CMake based linters the Python packages need linter tests instead: see https://github.com/ros2/demos/tree/f2a1d9f5fe694a11fb42f3c036dacd130141ac0e/demo_nodes_py/test

The information in the `setup.py` files should match the existing information in the `package.xml` files: e.g. description, license, authors.

I haven't checked the entry points myself but it looks weird that they reference a `main` function in the `rqt_gui.main` module which doesn't exist - only a `Main` and a check for `__main__`?

dirk-thomas:
Was this removed from the target branch after being merged?

brawner:
@dirk-thomas It's been reopened as #170. For an unknown reason Mike can't build the package as it is.

PR #170, opened 2018-12-08
Moving rqt to pure python (#169)

mlautman:
re-opening #169


brawner:
@dirk-thomas Addressing your comments from https://github.com/ros-visualization/rqt/pull/169#issuecomment-445437003

> The Python packages need to register themselves at the resource index as well as install their manifest file: see https://github.com/ros2/demos/blob/f2a1d9f5fe694a11fb42f3c036dacd130141ac0e/demo_nodes_py/setup.py#L10-L14 for an example.

Addressed

> Also since the patch removed all CMake based linters the Python packages need linter tests instead: see https://github.com/ros2/demos/tree/f2a1d9f5fe694a11fb42f3c036dacd130141ac0e/demo_nodes_py/test

Test files were added to rqt_gui and rqt_gui_py

> I haven't checked the entry points myself but it looks weird that they reference a main function in the rqt_gui.main module which doesn't exist - only a Main and a check for __main__?

A main function was added to main.py. I couldn't figure out how to get entry_points to call a script without a function as its entry point, nor could I figure out how to call the Main class method main.
https://github.com/ros-visualization/rqt/blob/0e54f9fa6af3565fad548bd11dbbd425a8f70881/rqt_gui/src/rqt_gui/main.py#L86

> The information in the setup.py files should match the existing information in the package.xml files: e.g. description, license, authors.

Addressed

rqt_gui_py/package.xml
```diff
@@ -25,6 +23,6 @@
 
   <export>
     <rqt_gui plugin="${prefix}/plugin.xml"/>
```
dirk-thomas, line 25: Where is this file being installed?

brawner: Added to data_files

dirk-thomas:
I don't know what the recommended approach for setuptools is but currently it only list a single author (see [previous](https://github.com/ros-visualization/rqt/pull/169#issuecomment-445437003) comment).

brawner:
It looks like as many names and emails can be added in the author name field, email address field can only be one though.

dirk-thomas:
Also in ROS 1 `rqt` can be invoked without `rosrun` since it is on the `PATH`. That should be the same in ROS 2. Since there can be only a single install destination for scripts the `rqt` entry point should go to `bin` and the `rqt_gui` one can stay a script which is is installed without being an entry point.

rqt_gui/setup.cfg
```diff
@@ -0,0 +1,4 @@
+[develop]
+script-dir=$base/lib/rqt_gui
+[install]
+install-scripts=$base/lib/rqt_gui
```
dirk-thomas: This file should removed to place the `rqt` executable into `bin`.

rqt_gui/setup.py
```diff
@@ -0,0 +1,44 @@
...
+    ],
+    description=(
+        'rqt_gui provides the main to start an instance of the ROS integrated graphical user ' +
+        'interface provided by qt_gui.'
+    ),
+    license='BSD',
+    tests_require=['pytest'],
+    entry_points={
+        'console_scripts': [
+            'rqt_gui = rqt_gui.main:main',
```
dirk-thomas: We remove `rqt_gui` from the console scripts again (with the other change only `rqt` is an entry point and goes to `bin`). Optionally we can restore a Python script `rqt_gui` and simply copy it to `lib/rqt_gui` (even if that doesn't make it `ros2 run`-able on Windows).

brawner: rqt is now a console script that is installed to bin/ and rqt_gui is a python script that gets copied over to lib/rqt_gui

brawner:
Tested on windows, and it has the expected effects. ```rqt``` can be run directly as a command, but ```ros2 run rqt_gui rqt_gui``` fails

brawner:
Linux CI https://ci.ros2.org/job/ci_linux/5859

dirk-thomas:
@ros-pull-request-builder retest this please

dirk-thomas:
@ros-pull-request-builder retest this please

rqt_gui_py/setup.py
```diff
@@ -0,0 +1,33 @@
...
+setup(
+    name=package_name,
+    version='0.5.0',
+    package_dir={'': 'src'},
+    packages=['rqt_gui_py'],
+    data_files=[
+        ('share/ament_index/resource_index/packages',
+            ['resource/' + package_name]),
+        ('share/' + package_name, ['package.xml']),
+        ('share/' + package_name + '/resource/', ['plugin.xml'])
```
dirk-thomas, line 13: See #179.

PR #174, opened 2018-12-11
Adding test dependency

brawner:
One of the recently added tests depends on std_srvs. This updates the package.xml to include it.

### DIFF
--- a/rqt_py_common/package.xml
+++ b/rqt_py_common/package.xml
@@ -30,6 +30,7 @@
   <exec_depend>rosidl_default_runtime</exec_depend>
 
   <test_depend>ament_cmake_pytest</test_depend>
+  <test_depend>std_srvs</test_depend>
 
   <member_of_group>rosidl_interface_packages</member_of_group>
 


dirk-thomas:
Instead of adding this test dependency can you just check the content of the service definition in that package?

brawner:
@dirk-thomas You mean just checking in the method whether std_srvs.srv.SetBool is actually imported?

This came up because I did a ci build up to rqt_py_common and std_srvs wasn't built. 
https://ci.ros2.org/job/ci_linux/5859/console

```
def test_get_service_text_from_class(self):  # noqa: D102
02:25:17 1:         from rqt_py_common.message_helpers import get_service_text_from_class
02:25:17 1: >       from std_srvs.srv import SetBool
02:25:17 1: E       ModuleNotFoundError: No module named 'std_srvs'
```

dirk-thomas:
> You mean just checking in the method whether std_srvs.srv.SetBool is actually imported?

No, the current test checks the return value of `get_service_text_from_class` for the service `std_srvs/SetBool`: https://github.com/ros-visualization/rqt/blob/3f0f2054bbff23e88b1d24a00171014e238d8c45/rqt_py_common/test/test_rqt_message_helpers.py#L65-L72

I am suggesting to check the content of the service file present in the test folder of that package: https://github.com/ros-visualization/rqt/blob/3f0f2054bbff23e88b1d24a00171014e238d8c45/rqt_py_common/test/srv/AddTwoInts.srv That would avoid the need for an external dependency.

dirk-thomas:
Superseded by #175.

Issue #181, opened 2018-12-13
ros2: Destroying subscriptions from a separate thread crashes python rqt_gui


brawner:
There is something not thread safe about creating a subscription in a rclpy plugin and then destroying it later in the same plugin. This was the stacktrace when python crashed.

This is currently reproducible in rqt_plot by subscribing to a topic and then deleting the subscriber. 
```
Thread 7 Crashed:: RclpySpinner
0   librmw_fastrtps_shared_cpp.dylib	0x000000011c06747e SubListener::hasData() + 126
1   librmw_fastrtps_shared_cpp.dylib	0x000000011c067213 check_wait_set_for_data(rmw_subscriptions_t const*, rmw_guard_conditions_t const*, rmw_services_t const*, rmw_clients_t const*) + 115
2   librmw_fastrtps_shared_cpp.dylib	0x000000011c068a97 rmw_fastrtps_shared_cpp::__rmw_wait(rmw_subscriptions_t*, rmw_guard_conditions_t*, rmw_services_t*, rmw_clients_t*, rmw_wait_set_t*, rmw_time_t const*)::$_0::operator()() const + 39
3   librmw_fastrtps_shared_cpp.dylib	0x000000011c068b0f bool std::__1::condition_variable::wait_until<std::__1::chrono::steady_clock, std::__1::chrono::duration<long long, std::__1::ratio<1l, 1000000000l> >, rmw_fastrtps_shared_cpp::__rmw_wait(rmw_subscriptions_t*, rmw_guard_conditions_t*, rmw_services_t*, rmw_clients_t*, rmw_wait_set_t*, rmw_time_t const*)::$_0>(std::__1::unique_lock<std::__1::mutex>&, std::__1::chrono::time_point<std::__1::chrono::steady_clock, std::__1::chrono::duration<long long, std::__1::ratio<1l, 1000000000l> > > const&, rmw_fastrtps_shared_cpp::__rmw_wait(rmw_subscriptions_t*, rmw_guard_conditions_t*, rmw_services_t*, rmw_clients_t*, rmw_wait_set_t*, rmw_time_t const*)::$_0) + 95
4   librmw_fastrtps_shared_cpp.dylib	0x000000011c068051 rmw_fastrtps_shared_cpp::__rmw_wait(rmw_subscriptions_t*, rmw_guard_conditions_t*, rmw_services_t*, rmw_clients_t*, rmw_wait_set_t*, rmw_time_t const*) + 2561
[...]

PR #184, opened 2018-12-15
Destroy listeners outside of spin_once()

brawner:
Destroying subscriptions is not thread-safe. Because the executor spins the node in a separate QThread outside of the plugins UI threads, destroying listeners needs to happen after spin_once().

This adds a method for plugins to call that registers subscriptions and clients for destruction. Unfortunately this also means that the spinner object needs to be passed around.

This address issue #181

### DIFF
--- a/rqt_gui_py/src/rqt_gui_py/rclpy_spinner.py
+++ b/rqt_gui_py/src/rqt_gui_py/rclpy_spinner.py
@@ -12,9 +12,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from queue import Queue
+
 from python_qt_binding.QtCore import qDebug, QThread, qWarning
 import rclpy
+from rclpy.client import Client
 from rclpy.executors import MultiThreadedExecutor
+from rclpy.subscription import Subscription
 
 
 class RclpySpinner(QThread):
@@ -23,6 +27,7 @@ class RclpySpinner(QThread):
         super().__init__()
         self._node = node
         self._abort = False
+        self._listener_queue = Queue()
 
     def run(self):
         qDebug('Start called on RclpySpinner, spinning ros2 node')
@@ -30,6 +35,12 @@ class RclpySpinner(QThread):
         executor.add_node(self._node)
         while rclpy.ok() and not self._abort:
             executor.spin_once(timeout_sec=1.0)
+
+            # Remove subscribers and clients that have been registered for destruction
+            while not self._listener_queue.empty():
+                listener = self._listener_queue.get()
+                self.__destroy_listener__(listener)
+
         if not self._abort:
             qWarning('rclpy.shutdown() was called before QThread.quit()')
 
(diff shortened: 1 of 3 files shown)


brawner:
Here is an example using it:

https://github.com/ros-visualization/rqt_plot/pull/20

brawner:
@dirk-thomas @mlautman Ping for a review?

rqt_gui_py/src/rqt_gui_py/rclpy_spinner.py
```diff
@@ -23,17 +27,56 @@ def __init__(self, node):
...
+        """
+        for listener in listeners:
+            if not isinstance(listener, Subscription) and not isinstance(listener, Client):
+                qWarning('Invalid object of type "{}" called for listener destruction'.format(
+                         type(listener)))
+                continue
+            qDebug('Registering destruction for listener of type "{}"'.format(type(listener)))
+            self._listener_queue.put(listener)
+
+    def __destroy_listener__(self, listener):
```
mlautman, line 69: Why this style? why not `_destroy_listener`?

mlautman (commented):
+1 after addressing 1 small comment

rqt_gui_py/src/rqt_gui_py/rclpy_spinner.py
```diff
@@ -23,17 +27,56 @@ def __init__(self, node):
...
+
         if not self._abort:
             qWarning('rclpy.shutdown() was called before QThread.quit()')
 
     def quit(self):  # noqa: A003
         qDebug('Quit called on RclpySpinner')
         self._abort = True
         super().quit()
+
+    def register_listeners_for_destruction(self, *listeners):
```
mlautman, line 52: This needs very prominent documentation. It is very reasonable for users to remove subscribers themselves if they don't know about this.

rqt_gui_py/src/rqt_gui_py/rclpy_spinner.py
brawner: Because external functions should absolutely not call this.

rqt_gui_py/src/rqt_gui_py/rclpy_spinner.py
brawner: Any suggestions? I imagine they will find rqt_gui crashing look at how other plugins do this

rqt_gui_py/src/rqt_gui_py/rclpy_spinner.py
mlautman: > Prepending a single underscore (_) has some support for protecting module variables and functions (not included with from module import *). While prepending a double underscore (__ aka “dunder”) to an instance variable or method effectively makes the variable or method private to its class (using name mangling) we discourage its use as it impacts readability and testability and isn’t really private.

The trailing double underscores could go

rqt_gui_py/src/rqt_gui_py/rclpy_spinner.py
mlautman: I really don't know. @dirk-thomas any ideas here? The issue we are having is that the spin method isn't thread safe and we are seeing segfaults when we try to unsubscribe while the executor is running.

mlautman: Correction, spin is thread safe but destroying subscribers is not.

brawner:
@dirk-thomas Any input on this PR and the accompanying rqt_plot PR?

https://github.com/ros-visualization/rqt_plot/pull/20

brawner:
rqt_topic also depends on this PR

dirk-thomas:
I don't think `rqt` should work around this by providing a difficult to discover API. Any Python application will have the same problem.

I think it would be better to improve the `rclpy` API to support this. Afaik `rclcpp` does allow you to use the API in that way.

brawner:
It doesn't look like there has been any action on the [related issue in rclpy](https://github.com/ros2/rclpy/issues/255) I created. I can investigate the bug in rclpy and work on a PR.

dirk-thomas:
> I can investigate the bug in rclpy and work on a PR.

That would be great!

brawner:
Unfortunately, unless I had some very specific guidance, I don't think I can resolve the related Rclpy issue within our current rqt effort. It looks like it's related to this issue:

https://github.com/ros2/rclpy/issues/192

https://github.com/brawner/rclpy/blob/test_destruction_outside_thread/rclpy/rclpy/executors.py#L29

mlautman:
Rather than wait indefinitely for the underlying bug to be fixed I think we should go ahead and merge this. Be sure to add comments explaining the spinner and noting that it can be removed once the related issues are closed

dirk-thomas:
We shouldn't merge this since it introduces API plugins would need to start using which they shouldn't need to once the`rclpy` API has been update.

dirk-thomas:
The upstream `rclpy` ticket has been resolved. So this should be obsolete.

PR #188, opened 2019-01-31
fixing srv and msg filters

mlautman:
This fixes the filtering of rosidl types. It is necessary for rqt_action to work as expected

### DIFF
--- a/rqt_py_common/src/rqt_py_common/message_helpers.py
+++ b/rqt_py_common/src/rqt_py_common/message_helpers.py
@@ -41,8 +41,8 @@ SRV_MODE = 'srv'
 ACTION_MODE = 'action'
 
 ROSIDL_FILTERS = {
-    MSG_MODE: lambda n: n.endswith('.msg'),
-    SRV_MODE: lambda n: n.endswith('.srv'),
+    MSG_MODE: lambda n: n.startswith('msg/') and n.endswith('.msg'),
+    SRV_MODE: lambda n: n.startswith('srv/') and n.endswith('.srv'),
     ACTION_MODE: lambda n: n.startswith('action/') and n.endswith('.idl')
 }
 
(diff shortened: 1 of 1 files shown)


rqt_py_common/src/rqt_py_common/message_helpers.py
```diff
@@ -227,7 +227,7 @@ def _get_rosidl_class_helper(message_type, mode, logger=None):  # noqa: C901
         return None
 
     message_info = message_type.split('/')
-    if len(message_info) != 2:
+    if len(message_info) not in [2, 3]:
```
dirk-thomas: In which case does the `message_info` have three components?

brawner: For services and actions, we add subfields to separate lines because there isn't currently a way of introspecting a service or action class like there is a message class. So in rqt_action for example, the first line will point to:

```
test_msgs/Fibonacci/Goal
```
Which is a rosidl msg.

brawner (approved):

rqt_py_common/src/rqt_py_common/message_helpers.py
dirk-thomas: These type of names are strictly defined in ROS (1): e.g. http://wiki.ros.org/Services

In this case `<pkgname>/<msgname>` or `<pkgname>/<srvname>`. There exists no such definition for "subfields" so I don't think it is a good idea to just come up with this pattern in this package.

Either this needs to be officially defined or (what I think would make more sense) add the missing introspection information to not require this in the first place.

dirk-thomas (dismissed):
Based on my inline comment.

rqt_py_common/src/rqt_py_common/message_helpers.py
mlautman: I removed this change

dirk-thomas (approved):

PR #189, opened 2019-02-11
[Windows][kinetic] build break fixes for roscpp_plugin_provider.cpp (adapted from windows_fixes)

seanyen:
- Cherry-pick changes from https://github.com/ros-visualization/rqt/commit/ea83b6a855e141e38bfefc293c569920bcd48076
- Only pick the build break fixes for roscpp_plugin_provider.cpp. Others seem ROS2 changes, so I skip them.

(This is verified in https://aka.ms/ros project.)


dirk-thomas:
Thanks for the patch.

Issue #198, opened 2019-05-29
[ROS2] Traceback when closing/exiting


Karsten1987:
When using `rqt_publisher` under dashing, I get a traceback when closing the application on OSX. This might be related to breaking changes in dashing related to passing the context object to `rcl_init()`. 
See here: https://github.com/ros2/rcl/blob/master/rcl/include/rcl/init.h#L82 and the related PR: https://github.com/ros2/rcl/pull/336

```
Traceback (most recent call last):
  File "/Users/karsten/workspace/osrf/ros2_full/build/rqt_gui_py/src/rqt_gui_py/rclpy_spinner.py", line 32, in run
    executor.spin_once(timeout_sec=1.0)
  File "/Users/karsten/workspace/osrf/ros2_full/install/lib/python3.7/site-packages/rclpy/executors.py", line 695, in spin_once
    handler, entity, node = self.wait_for_ready_callbacks(timeout_sec=timeout_sec)
  File "/Users/karsten/workspace/osrf/ros2_full/install/lib/python3.7/site-packages/rclpy/executors.py", line 649, in wait_for_ready_callbacks
    return next(self._cb_iter)
  File "/Users/karsten/workspace/osrf/ros2_full/install/lib/python3.7/site-packages/rclpy/executors.py", line 532, in _wait_for_ready_callbacks
    self._context.handle)
RuntimeError: Failed to initialize wait set: the given context is not valid, either rcl_init() was not called or rcl_shutdown() was called., at /Users/karsten/workspace/osrf/ros2_full/src/ros2/rcl/rcl/src/rcl/wait.c:130
```

dirk-thomas:
Does this only happen on macOS? Can you please try it on Ubuntu too. I can't reproduce the problem by just starting it with `ros2 run rqt_publisher rqt_publisher` and closing it through the [X] button.

Karsten1987:
I currently don't have access to a graphical Ubuntu.
It could definitely just be a OSX problem - yet I can reproduce it in 100% of the cases.

Karsten1987:
So I've tried it on a 18.04 docker instance and see the same error message when closing the app.

dirk-thomas:
Similar to ros2/demos#316.

dirk-thomas:
@Karsten1987 This seems to be a problem outside of `rqt`, I tried to reproduce the problem on Foxy but the error doesn't show up. Can you check if the problem is still happening for you in the latest distro Foxy.

[...]

PR #199, opened 2019-05-29
suppress traceback when ctrl-c

Karsten1987:
fixes ros-visualization/rqt#198

unfortunately, rclpy doesn't throw a more specific error message than `RuntimeError`. In order to narrow down the reasons for the error, we parse the error message to hit the specific keywords. 

Signed-off-by: Karsten Knese <karsten@openrobotics.org>

### DIFF
--- a/rqt_gui_py/src/rqt_gui_py/rclpy_spinner.py
+++ b/rqt_gui_py/src/rqt_gui_py/rclpy_spinner.py
@@ -29,10 +29,18 @@ class RclpySpinner(QThread):
         executor = MultiThreadedExecutor()
         executor.add_node(self._node)
         while rclpy.ok() and not self._abort:
+            # we have to put a try/except block around the executor
+            # because when ctrl-c there might be the case of waiting within the executor
+            # even though rcl shutdown() was triggered already.
+            # see ticket here: https://github.com/ros-visualization/rqt/issues/198
             try:
                 executor.spin_once(timeout_sec=1.0)
-            except RuntimeError:
-                if rclpy.ok():
+            except RuntimeError as re:
+                # right now there is no specific error thrown other than RuntimeError
+                # in order to narrow down the reason for the error we parse the error message
+                if not rclpy.ok() and 'Failed to initialize wait set' in str(re):
+                    pass
+                else:
                     raise
         if not self._abort:
             qWarning('rclpy.shutdown() was called before QThread.quit()')
--- a/rqt_gui_py/src/rqt_gui_py/rclpy_spinner.py
+++ b/rqt_gui_py/src/rqt_gui_py/rclpy_spinner.py
@@ -29,7 +29,11 @@ class RclpySpinner(QThread):
         executor = MultiThreadedExecutor()
         executor.add_node(self._node)
         while rclpy.ok() and not self._abort:
-            executor.spin_once(timeout_sec=1.0)
+            try:
+                executor.spin_once(timeout_sec=1.0)
+            except RuntimeError:
+                if rclpy.ok():
+                    raise
         if not self._abort:
             qWarning('rclpy.shutdown() was called before QThread.quit()')
 


Karsten1987:
This is up for review.

sloretz:
> unfortunately, rclpy doesn't throw a more specific error message than RuntimeError. 

I would expect `KeyboardInterrupt` on `ctrl-c`, not `RuntimeError`

Karsten1987:
Can you reproduce this? Whenever I close the GUI or hit ctrl-c I get the same traceback, only mentioning a RuntimeError.

dirk-thomas:
Please post the exact traceback you are seeing.

Karsten1987:
Traceback was posted in the referenced issue:
```
Traceback (most recent call last):
  File "/Users/karsten/workspace/osrf/ros2_full/build/rqt_gui_py/src/rqt_gui_py/rclpy_spinner.py", line 32, in run
    executor.spin_once(timeout_sec=1.0)
  File "/Users/karsten/workspace/osrf/ros2_full/install/lib/python3.7/site-packages/rclpy/executors.py", line 695, in spin_once
    handler, entity, node = self.wait_for_ready_callbacks(timeout_sec=timeout_sec)
  File "/Users/karsten/workspace/osrf/ros2_full/install/lib/python3.7/site-packages/rclpy/executors.py", line 649, in wait_for_ready_callbacks
    return next(self._cb_iter)
  File "/Users/karsten/workspace/osrf/ros2_full/install/lib/python3.7/site-packages/rclpy/executors.py", line 532, in _wait_for_ready_callbacks
    self._context.handle)
RuntimeError: Failed to initialize wait set: the given context is not valid, either rcl_init() was not called or rcl_shutdown() was called., at /Users/karsten/workspace/osrf/ros2_full/src/ros2/rcl/rcl/src/rcl/wait.c:130
```

dirk-thomas:
This exception should not be catched and ignored but actually be resolved so that it isn't raised in the first place.

dirk-thomas:
I just noticed that you pushed changes to this branch. Without commenting on the ticket there is no notification so it is unlikely that I look at the ticket again.

The current patch checks for a specific exception string which seems pretty hacky. I don't think this should be merged as is. Instead this should be addressed in `rclpy` and if there is a need to handle a custom exception here the patch can be updated.

Karsten1987:
I believe I just rebased that branch and didn't really put on new code for it. I can open a ticket on rclpy for it - not sure I have much time to address the issue on rclpy level though.

Karsten1987:
closing wrt https://github.com/ros2/rclpy/issues/398

rqt_gui_py/src/rqt_gui_py/rclpy_spinner.py
```diff
@@ -29,7 +29,19 @@ def run(self):
...
         executor.add_node(self._node)
         while rclpy.ok() and not self._abort:
-            executor.spin_once(timeout_sec=1.0)
+            # we have to put a try/except block around the executor
+            # because when ctrl-c there might be the case of waiting within the executor
+            # even though rcl shutdown() was triggered already.
+            # see ticket here: https://github.com/ros-visualization/rqt/issues/198
+            try:
+                executor.spin_once(timeout_sec=1.0)
+            except RuntimeError as re:
```
ivanpauno, line 38: I'm a starting to dig in `rclpy` signal handling related problems.
Could you try the following?
```python3
try:
    executor.add_node(self._node)
    while executor.context.ok() and not self._abort:
        executor.spin_once()
finally:
    executor.remove_node(node)
```
(replacing lines 31-44 of the patch)

I don't have a mac available to try it out.

Karsten1987: unfortunately, this didn't help and resulted in the same traceback. Which makes sense as I believe the exception emerges from the `spin_once()` call.

ivanpauno: Ok. The error is reproducible in Ubuntu too, I will try to take a look.

Issue #197, opened 2019-05-29
Malformed message_type rqt ROS2


skucheria:
**- Operating System:**
Ubuntu 18.04
**- Installation type:**
from source
**- DDS implementation:**
RTI Connext
**- Client library (if applicable):**
rqt

Receiving "malformed message_type" error on message types in rqt when attempting to use rqt_plot or topic monitor. 
Ex:

```
[ERROR] [get_message_class]: Malformed message_type: std_msgs/msg/String
TopicInfo.__init__(): topic "/robot_description": can not get message class for type "std_msgs/msg/String"
```

dirk-thomas:
This should be resolved on the latest HEAD on the `crystal-devel` branch. Please try it again to confirm that it is fixed for you.

dirk-thomas:
I will go ahead and close this assuming the problem has already been resolved. Please feel free to comment on the closed ticket and if necessary it can be reopened.

PR #200, opened 2019-05-29
parse message type description correctly

Karsten1987:
Requires branching for dashing!

Fixes ros-visualization/rqt#197

Signed-off-by: Karsten Knese <karsten@openrobotics.org>


dirk-thomas:
> Requires branching for dashing!

Instead of breaking compatibility the patch could easily be updated to be compatible with both version. I will edit it in place...

PR #206, opened 2019-07-31
Use catkin_install_python for python script.

mikepurvis:
This aids in our Python 3 migration.


dirk-thomas:
Thanks for the patch.

PR #216, opened 2020-02-03
Allow definition of settings file

agutenkunst:
I would need this feature since I need to provide a settings file to the user.

Does not change default behavior.


dirk-thomas:
Thanks for the improvement.

dirk-thomas:
Cherry-picked to the `crystal-devel` branch: 408c9beccd987360547076451bbd63622c32f13a.

PR #228, opened 2020-06-12
 avoid installing test interfaces 

dirk-thomas:
Replaces #214.

The test in `test_rqt_roscomm_util.py` is testing stuff which the package doesn't explicitly depend on and was ROS 1 specific (see the comments in the file). So I just got rid of the test.


rqt_py_common/CMakeLists.txt
```diff
@@ -25,14 +25,30 @@ if(BUILD_TESTING)
...
+  file(WRITE "${CMAKE_CURRENT_BINARY_DIR}/ament_index/share/ament_index/resource_index/rosidl_interfaces/${_PROJECT_NAME}")
+  file(WRITE "${CMAKE_CURRENT_BINARY_DIR}/ament_index/share/ament_index/resource_index/rosidl_interfaces/${PROJECT_NAME}"
+    "msg/Val.msg\n"
+    "msg/ArrayVal.msg\n"
+    "srv/AddTwoInts.srv")
+  set(PROJECT_NAME ${_PROJECT_NAME})
 
   ament_add_pytest_test(${PROJECT_NAME} test
-    APPEND_ENV PYTHONPATH=${CMAKE_CURRENT_BINARY_DIR}
+    ENV "AMENT_PREFIX_PATH=${CMAKE_CURRENT_BINARY_DIR}/ament_index"
```
cottsay, line 50: Wouldn't using `APPEND_ENV` allow you to avoid creating fake markers for the real project?

cottsay (approved):
Thanks for coming up with a solution here.

rqt_py_common/CMakeLists.txt
dirk-thomas: Unfortunately no.

The code is written so poorly that it would fail if package `rqt_py_common` is in the index but doesn't install interfaces. Therefore appending the custom prefix path isn't sufficient since `rqt_py_common` would be found before. Atm there is no `PREPEND_ENV` option which would at least avoid the second problem.

cottsay: Ah, unfortunate.

In that case, is this an option?
```
ENV "AMENT_PREFIX_PATH=${CMAKE_CURRENT_BINARY_DIR}/ament_index;$ENV{AMENT_PREFIX_PATH}"
```

dirk-thomas: Yes, it just results in a very large command for the test invocation. That is why I scaled it back to only what is needed since the tests don't mind about the other prefix path.

cottsay: Thanks for explaining.

dirk-thomas: Np, the fact that we need this kind of workaround is already unfortunate...

PR #231, opened 2020-07-06
Use spin instead of spin_once

DensoADAS:
Only spin() executes the callbacks multithreaded. Using spin_once() the multithreaded executor has no effect.


kfabian:
Is there something unclear regarding this PR or is the single-threaded execution intended?

ahcorde:
This PR is targeting an [EOL ROS distro](https://docs.ros.org/en/rolling/Releases/End-of-Life.html). Closing this. Feel free to reopen the PR against `rolling`

PR #241, opened 2020-12-16
Changed getiter to iter

sven-herrmann:
Replaced the function call getiter with the proposed replacement iter according to the [Python docs ](https://docs.python.org/3.8/library/xml.etree.elementtree.html#xml.etree.ElementTree.ElementTree.getiterator).


timonegk:
@mjeronimo Could this be merged soon? It is really a trivial fix and doesn't break anything (`getiterator` has been deprecated since python 3.2 / 2.7).

mjeronimo (approved):

PR #248, opened 2021-04-05
Fix a crash at shutdown

mjeronimo:
Join (QThread.wait) the spinner thread before continuing with the shutdown.

Signed-off-by: Michael Jeronimo <michael.jeronimo@openrobotics.org>

### DIFF
--- a/rqt_gui_py/src/rqt_gui_py/ros_py_plugin_provider.py
+++ b/rqt_gui_py/src/rqt_gui_py/ros_py_plugin_provider.py
@@ -30,7 +30,7 @@
 
 import os
 
-from python_qt_binding.QtCore import qDebug
+from python_qt_binding.QtCore import qDebug, qWarning
 from qt_gui.composite_plugin_provider import CompositePluginProvider
 
 import rclpy
@@ -54,7 +54,10 @@ class RosPyPluginProvider(CompositePluginProvider):
         qDebug('Shutting down RosPyPluginProvider')
         if self._spinner:
             self._spinner.quit()
-            self._spinner.wait(msecs=self._shutdown_timeout)
+            joined = self._spinner.wait(msecs=self._shutdown_timeout)
+            if not joined:
+                qWarning('Timed out attempting to join the RclpySpinner thread')
+                return
         if self._node:
             self._destroy_node()
         super().shutdown()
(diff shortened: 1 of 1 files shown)


rqt_gui_py/src/rqt_gui_py/ros_py_plugin_provider.py
```diff
@@ -48,11 +48,13 @@ def __init__(self):
         self._node_initialized = False
         self._node = None
         self._spinner = None
+        self._shutdown_timeout = 2000
 
     def shutdown(self):
         qDebug('Shutting down RosPyPluginProvider')
         if self._spinner:
             self._spinner.quit()
+            self._spinner.wait(msecs=self._shutdown_timeout)
```
ivanpauno: If `wait()` timed out, cleaning up the node will still be an issue.
Maybe something like:
```suggestion
            joined = self._spinner.wait(msecs=self._shutdown_timeout)
        if not joined:
            # maybe log an error
            return
```

The other option is to not pass a timeout, and block forever until joining the thread.

mjeronimo: No option is ideal:

* If there is a return after the the wait and the wait times out, the app will exit, but the node won't get cleaned up and will still appear when doing "ros2 node list." 
* If the wait times out and there is no return statement (the code continues, cleaning up the node), there is still a risk of a crash (although slight). If there is a crash, the node is not cleaned up properly and still appears when doing "ros2 node list"
* If waiting without a timeout, if there is an issue with the spinner thread, the app will hang

I suppose from the user's perspective, the first option would be the best. At least there wouldn't be a crash or hang. The node would still appear in the list for a while. What do you think?

ivanpauno: > I suppose from the user's perspective, the first option would be the best. At least there wouldn't be a crash or hang. The node would still appear in the list for a while. What do you think?

:+1: the first option sounds reasonble to me.
The node will dissappear after a time of `ros2 node list` too, so that's not a big issue.

ivanpauno (approved):

mjeronimo:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14274/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9064/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11958/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14374/)

PR #250, opened 2021-04-22
Fix up the package description.

clalancette:
It described a set of metapackages that we don't actually have.
Just remove most of it here and say that RQt is a Qt-based
GUI development system for ROS.

Signed-off-by: Chris Lalancette <clalancette@openrobotics.org>


ivanpauno (approved):

clalancette:
Since this is just a change to the package.xml, I'm going to merge it without further CI.  Thanks for the review!

PR #269, opened 2022-05-11
Updating maintainer

quarkytale:
Added new maintainer and moved previous as authors!


quarkytale:
@clalancette should this be on humble branch instead?

clalancette (approved):
This change is fine, but you should also update the `ros2` branch here.

PR #282, opened 2022-10-24
Add in 3-clause BSD license file.

### LINKED ISSUE #281: Missing License file
I know this is a ROS package and it uses the same license as the rest of ROS (which I think is Apache V2).

However if someone that doesn't know that arrive to this GitHub page, they will not know under which license this project is.

clalancette:
Signed-off-by: Chris Lalancette <clalancette@openrobotics.org>

This will fix #281


mjcarroll (approved):

clalancette:
Merging without CI since it won't make a difference here.

PR #289, opened 2023-04-03
Changes the use of __slots__ for the field and field type getter

Voldivh:
This PR modifies the use of `__slots__` for the appropriate message components name _getter_. This changes should solve the problems when [this PR](https://github.com/ros2/rosidl_python/pull/194) is incorporated.

### DIFF
--- a/rqt_py_common/src/rqt_py_common/topic_dict.py
+++ b/rqt_py_common/src/rqt_py_common/topic_dict.py
@@ -68,8 +68,8 @@ class TopicDict(object):
             'children': {},
         }
 
-        if hasattr(field, '__slots__'):
-            for child_slot_name in field.__slots__:
+        if hasattr(field, '_fields_and_field_types'):
+            for child_slot_name in field.get_fields_and_field_types().keys():
                 field_dict[slot_name]['children'].update(
                     self._recursive_create_field_dict(
                         child_slot_name, getattr(field, child_slot_name)))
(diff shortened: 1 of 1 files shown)


rqt_py_common/src/rqt_py_common/topic_dict.py
```diff
@@ -68,8 +68,8 @@ def _recursive_create_field_dict(self, slot_name, field):
             'children': {},
         }
 
-        if hasattr(field, '__slots__'):
-            for child_slot_name in field.__slots__:
+        if hasattr(field, '_fields_and_field_types'):
+            for child_slot_name in field.get_fields_and_field_types().keys():
```
clalancette: This isn't a slot name anymore; it is a field name.  So I think we should rename this field to `field_name` (same below where we use it).

clalancette (changes requested):
One change inline, otherwise this looks good to me.

rqt_py_common/src/rqt_py_common/topic_dict.py
Voldivh: Makes sense, I'll take that into account with the other PR's that solve similar issues.

Voldivh: Done.

clalancette (approved):
Looks good to me, thanks!

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18397/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12943/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19078/)

Issue #270, opened 2022-05-13
`ros2 run rqt rqt` No Executable Found


sloretz:
```
C:\Users\IEUser>ros2 run rqt rqt
No executable found
```

`rqt` works fine, but `ros2 run rqt rqt` does not. For that to work, the executable needs to be installed twice like `rviz` does

https://github.com/ros2/rviz/blob/7b4ba855b5803ef8b77b7732889c8b1b6e43532b/rviz2/CMakeLists.txt#L82-L83

I don't know how to do that with a `setup.py`. Most ROS 2 packages use the `setup.cfg` to change the install location, but it's not obvious to me how to duplicate that.

https://github.com/ros-visualization/rqt/blob/302a2ea7f9a2f85c4e2d11d793419d576d47fa15/rqt_gui/setup.py#L41

Voldivh:
After taking a closer look, I just noticed that the correct command we have to use is `ros2 run rqt_gui rqt_gui` and we get the exact same behavior as when we only run the entry point `rqt`. I'm not sure if this is the expected behavior, I think it does because on the package `rqt` we don't have any executable or UI or anything, it is basically description for the package.

Voldivh:
Let me know what you think @clalancette @sloretz @Yadunund

clalancette:
> After taking a closer look, I just noticed that the correct command we have to use is `ros2 run rqt_gui rqt_gui` and we get the exact same behavior as when we only run the entry point `rqt`. I'm not sure if this is the expected behavior, I think it does because on the package `rqt` we don't have any executable or UI or anything, it is basically description for the package.

If I take a look at https://github.com/ros-visualization/rqt/blob/95d3d6e4335ad93244fb4063ce2fa2fe9e80e532/rqt_gui/setup.py, then I see two things:

```
    data_files=[
        ('share/' + package_name + '/resource', [
            'resource/rqt.graffle',
            'resource/rqt.png',
            'resource/rqt.svg',
        ]),
        ('share/ament_index/resource_index/packages',
            ['resource/' + package_name]),
        ('share/' + package_name, ['package.xml']),
        ('lib/' + package_name, ['bin/rqt_gui'])
    ],
```

and

```
    entry_points={
        'console_scripts': [
            'rqt = rqt_gui.main:main',
        ],
    },
```

[...]

PR #293, opened 2023-05-16
Modifies the name of rqt_gui to rqt

Voldivh:
This PR addresses the issue https://github.com/ros-visualization/rqt/issues/270.


clalancette:
So I learned a little more about this yesterday.

In particular, this script cannot ever work on Windows, because Windows does not support running a python script directly.  On Windows, you *must* install it via a console_script, which is why `rqt` works.

I'm honestly not sure what we want this to look like going forward.  I've put an item on the agenda for the next ROS 2 meeting to discuss this.

PR #298, opened 2023-12-12
Git ignore

mosfet80:
Fix to ignore local .cproject and .project push


ahcorde (changes requested):
I don't know if this PR is still relevant, but are you sure you want to target `kinetic` ?

PR #304, opened 2024-02-15
Stop using python_cmake_module.

clalancette:
We really don't need it anymore, and can just use the builtin find_package(Python3).

This must be merged before https://github.com/ros2/ros2/pull/1524 ; see that pull request for more information about this change.

### DIFF
--- a/rqt_py_common/CMakeLists.txt
+++ b/rqt_py_common/CMakeLists.txt
@@ -18,15 +18,6 @@ if(BUILD_TESTING)
 
   find_package(rosidl_default_generators REQUIRED)
   find_package(ament_cmake_pytest REQUIRED)
-  # Provides PYTHON_EXECUTABLE_DEBUG
-  find_package(python_cmake_module REQUIRED)
-  find_package(PythonExtra MODULE)
-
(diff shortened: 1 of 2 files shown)


quarkytale:
This is also not calling `find_package(Python3)` should `ament_cmake_python` be added then?

clalancette:
> This is also not calling `find_package(Python3)`

Yes, that's correct, because there is no direct reference to `Python3_EXECUTABLE` in this CMakeLists.txt

> should `ament_cmake_python` be added then?

But you are right about this.  Since this CMakeLists.txt is calling [ament_python_install_package](https://github.com/ros-visualization/rqt/blob/51aed01e4b83ac58c2684346a3945d1d56f6a8f1/rqt_py_common/CMakeLists.txt#L8), it should declare a dependency on `ament_cmake_python`.  I'll add one to this PR.

quarkytale (approved):

clalancette:
CI for this (along with all of the rest of the changes needed) is in https://github.com/ros2/ros2/pull/1524#issuecomment-2365359588

PR #312, opened 2024-07-24
Use an rclpy context manager.

clalancette:
This makes the code cleaner and ensures we cleanup no matter how we exit.

### DIFF
--- a/rqt_py_common/src/rqt_py_common/topic_completer.py
+++ b/rqt_py_common/src/rqt_py_common/topic_completer.py
@@ -95,52 +95,54 @@ if __name__ == '__main__':
         QTreeView, QVBoxLayout, QWidget
 
     import rclpy
-    rclpy.init()
-    topic_completer_node = rclpy.create_node()
+    from rclpy.executors import ExternalShutdownException
 
-    app = QApplication(sys.argv)
-    mw = QMainWindow()
-    widget = QWidget(mw)
-    layout = QVBoxLayout(widget)
+    try:
+        with rclpy.init():
+            topic_completer_node = rclpy.create_node()
 
-    edit = QLineEdit()
-    edit_completer = TopicCompleter(edit)
-    edit_completer.update_topics(topic_completer_node)
-    # edit_completer.setCompletionMode(QCompleter.InlineCompletion)
-    edit.setCompleter(edit_completer)
+            app = QApplication(sys.argv)
+            mw = QMainWindow()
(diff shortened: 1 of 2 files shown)


rqt_py_common/src/rqt_py_common/topic_completer.py
```diff
@@ -95,52 +95,54 @@ def update_topics(self, node):
...
-    mw.move(300, 0)
-    mw.resize(800, 900)
-    mw.show()
-    app.exec_()
-
-    topic_completer_node.destroy_node()
-    rclpy.shutdown()
+    from rclpy.executors import ExternalShutdownException
+
+    try:
```
sloretz, line 100: Nit on the scope of this try/except since it's a bigger chunk of code. I think the `try/except` only needs to surround the part that blocks, which IIUC is the `app.exec_()`. I'd recommend reducing it to just that in this case.

clalancette: > Nit on the scope of this try/except since it's a bigger chunk of code. I think the `try/except` only needs to surround the part that blocks, which IIUC is the `app.exec_()`. I'd recommend reducing it to just that in this case.

I've been opting for surrounding most/all of the code in the try/except block.  The basic reason for that is to capture the `KeyboardInterrupt` everywhere.  That is, if you happen to be in the "setup" phase (before the `app.exec_()`), and you hit Ctrl-C, then you would still get an ugly exit with the raise of the `KeyboardInterrupt` exception.  While this is fairly unlikely, by surrounding everything we make sure that we always have a "clean" exit.

That said, I don't feel that strongly about it, so if you think it is better to reduce the scope I can go ahead and do that.

sloretz: I don't feel strongly about it either. Nothing really bad happens from putting too much code into a try/except block that catches specific exceptions. It just hides what code is actually expected to raise

sloretz (approved):

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21622/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15909/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1126/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22339/)

adityapande-1995 (approved):

PR #320, opened 2025-04-12
Bump cmake_minimum_required to avoid deprecation

arne48:
Noetic will be EOL very soon and it was already discussed what last changes would be worthwhile.
This was done in [Discourse](https://discourse.ros.org/t/ros-noetic-end-of-life-may-31-2025/43160/4) .
One point was the cmake_minimum_required version because <3.5 is deprecated in newer CMake version.


sloretz (approved):
Thank you for the PR!

PR #346, opened 2026-04-28
Use the same Qt version as qt_gui_cpp

cottsay:
Rather than running the detection logic again, just use the version that qt_gui_cpp used, which we must match.


sloretz (approved):

mjcarroll (approved):

ahcorde (changes requested):
CMake defaults to ascending resolution, so when Qt5 and Qt6 are both present, Qt5 will be found first when Qt6 is desired (Rolling, L-Turtle, and beyond). e.g. this pattern find_package(QT NAMES Qt6 Qt5 ...) resolves to Qt5.

Reference https://github.com/ros2/rviz/pull/1689

